PR #29, opened 2026-02-21
feat(channels): telegram streaming

veelenga:
- telegram message streaming
- agent loop refactoring 
- specs


veelenga:
This is working and looking nice on telegram, but could be an overkill for this project for now.

PR #30, opened 2026-02-21
fix(agent): preserve `extra_content`, primarily for gemini

xendk:
Gemini requires `extra_content.google.thought_signature` to be retained on tool calls.

Google apparently decided on the nested structure so other providers might use it too.

### DIFF
--- a/src/autobot/providers/http_provider.cr
+++ b/src/autobot/providers/http_provider.cr
@@ -80,9 +80,7 @@ module Autobot
         apply_auth_headers(headers, spec)
 
         Log.debug { "POST #{url} model=#{model}" }
-        Log.debug { body.to_pretty_json }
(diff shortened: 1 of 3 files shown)


src/autobot/providers/http_provider.cr
```diff
@@ -80,7 +80,9 @@ module Autobot
         apply_auth_headers(headers, spec)
 
         Log.debug { "POST #{url} model=#{model}" }
+        Log.debug { body.to_pretty_json }
         response = http_post(url, headers, body.to_json)
+        Log.debug { response.body }
```
veelenga: this could be very noisy and potentially leak sensitive data (API keys in headers aren't logged, but request/response bodies may contain user content)

PR #35, opened 2026-02-22
feat(channels): add zulip

xendk:
[Zulip](https://zulip.com/) is a Slack alternative.

Rudimentary first implementation, supports direct messages.


xendk:
Ah haha, so *that's* why Gemini decided to fold out things into methods.

I'll get it fixed tomorrow.

PR #56, opened 2026-03-14
fix(cli): add workspace security warnings to doctor command

renich:
Updates the autobot doctor command to check for risky workspace locations. It specifically adds a warning if the workspace is initialized directly in the user's home directory, which can lead to unintentional file access or permission issues.

Changes:

- Added home directory check in src/autobot/cli/doctor.cr.
- Improved output formatting for security warnings.


veelenga (approved):
lgtm

PR #58, opened 2026-03-14
feat(tools): allow overriding exec safety guards via config

renich:
Adds the ability for users to explicitly override default execution safety guards through the config.yml file. This allows for more permissive operations in sandboxed or trusted environments.

Changes:

- Added safety_override flag to tool configuration.
- Updated ExecTool to respect user-defined safety bypasses.

### DIFF
--- a/src/autobot/cli/gateway.cr
+++ b/src/autobot/cli/gateway.cr
@@ -141,8 +141,7 @@ module Autobot
         cron_service : Cron::Service,
       )
         sandbox_config = config.tools.try(&.sandbox) || "auto"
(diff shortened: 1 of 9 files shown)


src/autobot/cli/gateway.cr
```diff
@@ -141,6 +141,8 @@ module Autobot
         cron_service : Cron::Service,
       )
         sandbox_config = config.tools.try(&.sandbox) || "auto"
+        deny_patterns = config.tools.try(&.exec.try(&.deny_patterns)).try(&.map { |pat| Regex.new(pat, Regex::Options::IGNORE_CASE) }) || Tools::ExecTool::DEFAULT_DENY_PATTERNS
```
veelenga: Config deny_patterns silently replaces all built-in safety defaults.

The `||` fallback only kicks in when the config key is absent. If a user provides `deny_patterns: []` (empty array) or even a single custom pattern, all 25+ built-in deny patterns (fork bombs, privilege escalation, reverse shells, etc.) are dropped.

PR #65, opened 2026-06-07
fix(tools): prevent deadlock when executed commands spawn daemons

renich:
## Description
When Autobot uses the `exec` tool, if the executed command spawns a background daemon (such as `postgres` via `pg_ctl start`), that daemon inherits the standard output and standard error file descriptors from the shell. 

Because Autobot's process executor was waiting indefinitely for the read channels to close (EOF), and the long-running background daemon never closed those write pipes, the agent's main event loop would hang permanently. 

## Fix
This PR ensures that the read ends of the `stdout` and `stderr` pipes are explicitly closed by the parent process immediately after the main process completes. This forcefully interrupts any lingering blocking reads from background fibers and safely returns the captured output up to that point.

### DIFF
--- a/src/autobot/tools/exec.cr
+++ b/src/autobot/tools/exec.cr
@@ -185,12 +185,14 @@ module Autobot
 
         timed_out, status = wait_for_process(process, completed)
 
+        # Close read ends to break any blocking io.read in background fibers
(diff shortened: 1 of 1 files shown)


veelenga (approved):
lgtm

PR #73, opened 2026-06-11
fix(tools): respect sandbox: none config in SandboxExecutor

renich:
This pull request ensures that filesystem and execution tools correctly bypass the sandboxing mechanism when `sandbox` is configured as `none`. Previously, SandboxExecutor would still attempt to sandbox operations if a workspace was configured.

### Changes:
- Added a `sandboxed` flag in `SandboxExecutor` to control whether sandboxing is applied.
- Updated all execution and filesystem operations in `SandboxExecutor` to fall back to direct executions when `sandboxed` is false.
- Initialized `SandboxExecutor` in `tools.cr` with the resolved sandbox setting (for both main and subagent registries).
- Added a unit spec `spec/autobot/tools/sandbox_executor_spec.cr` to verify that `SandboxExecutor` correctly bypasses the sandbox and runs direct operations when `sandboxed` is set to `false`.

### AI Assistance Disclosure
This pull request was prepared with the assistance of Gemini AI. The user (Rénich) has directed, reviewed, and tested these changes, and takes full responsibility for the code.

### DIFF
--- a/docs/sandboxing.md
+++ b/docs/sandboxing.md
@@ -227,16 +227,16 @@ autobot agent --sandbox docker
 
 ## Development
 
-### Running Without Sandbox (Tests)
+### Running Without Sandbox (Tests and Development)
 
-Tests automatically disable sandboxing:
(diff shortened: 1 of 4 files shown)


veelenga (approved):
lgtm

PR #83, opened 2026-07-27
hardening: SQLite tool bypassed exec denylist via dot-commands

renich:
🚨 **Severity:** CRITICAL
💡 **Vulnerability:** The SQLite plugin executed user-provided SQL queries using the `sqlite3` CLI without the `-safe` flag, allowing command injection via dot-commands like `.shell` or `.system`.
🎯 **Impact:** An untrusted query could execute arbitrary shell commands via the `.system` dot-command, leading to Remote Code Execution (RCE).
🔧 **Fix:** Appended the `-safe` flag to all `sqlite3` CLI invocations in `src/autobot/plugins/builtin/sqlite.cr`, disabling dangerous dot-commands.
✅ **Verification:** Ran `crystal spec` (974 passing) and `./bin/ameba` (0 failures).

### DIFF
--- a/src/autobot/plugins/builtin/sqlite.cr
+++ b/src/autobot/plugins/builtin/sqlite.cr
@@ -207,12 +207,12 @@ module Autobot
         private def init_migrations_table(db_name : String) : Tools::ToolResult
           sql = "CREATE TABLE IF NOT EXISTS schema_migrations " \
                 "(version TEXT PRIMARY KEY, applied_at DATETIME DEFAULT CURRENT_TIMESTAMP);"
-          @executor.exec("sqlite3 #{shell_escape(db_path(db_name))} #{shell_escape(sql)}", timeout: INIT_TIMEOUT)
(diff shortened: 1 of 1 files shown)


veelenga:
Two things pull it down from "critical" in the context of this project:

1. It runs inside the sandbox. SandboxExecutor is sandboxed by default (bubblewrap/Docker unless sandbox=none). A .system call executes inside that sandbox, so it's not a host-level compromise — it's confined to the same jail every tool already runs in.
2. The agent already has arbitrary shell. ExecTool (and BashTool) are registered by default and let the agent run shell commands directly. So .system via the SQLite tool doesn't grant a capability the agent lacks — its real significance is that it bypasses ExecTool's DEFAULT_DENY_PATTERNS (the denylist blocking rm -rf, curl|sh, sudo, reverse shells, etc.). A query tool that's supposed to be SQL-only becoming an unfiltered command channel is a legitimate problem, just not "RCE from nothing."

The realistic threat model is prompt injection: untrusted content (web fetch results, tool output) steers the agent into issuing a .system query, evading the exec denylist. That's worth fixing.

veelenga (approved):

PR #98, opened 2026-08-24
fix(agent): synchronize SubagentManager running tasks for multithreading safety

### LINKED ISSUE #97: SubagentManager unprotected @running_tasks hash mutation and missing exception cleanup
### Description
In `src/autobot/agent/subagent.cr`, `SubagentManager` maintains an in-memory hash `@running_tasks : Hash(String, Bool)` to track active background subagent tasks.

### Problem Details
1. **Unsynchronized Hash Mutation Across Fibers**:
   `@running_tasks` is written in caller fibers (`@running_tasks[task_id] = true`), deleted in background fibers (`@running_tasks.delete(task_id)`), and read via `running_count` without `Mutex` synchronization. Under multithreaded runtime (`-Dpreview_mt`), concurrent Hash mutations cause data races and state corruption.
2. **Missing `ensure` Cleanup on Fiber Exceptions**:
   In `spawn`:
   ```crystal
   ::spawn do
     run_subagent(task_id, task, display_label, origin)
     @running_tasks.delete(task_id)
   end
   ```
   If `run_subagent` encounters an unhandled exception, `@running_tasks.delete(task_id)` is bypassed, permanently leaking a zombie task in `running_count`.

### Proposed Fix
- Introduce `@tasks_mutex : Mutex = Mutex.new` to synchronize all reads, writes, and deletions of `@running_tasks`.
- Wrap `run_subagent` in `begin ... ensure @tasks_mutex.synchronize { @running_tasks.delete(task_id) } end`.

renich:
## Description
In `SubagentManager` (`src/autobot/agent/subagent.cr`), `@running_tasks : Hash(String, Bool)` tracks active background subagents.

Previously:
1. Writes (`spawn`), deletions (`::spawn`), and reads (`running_count`) on `@running_tasks` were executed concurrently without synchronization, leading to data races under multithreaded runtime (`-Dpreview_mt`).
2. `@running_tasks.delete(task_id)` was placed sequentially after `run_subagent` without an `ensure` block. If `run_subagent` encountered an unhandled error, the task entry was never deleted and leaked a permanent zombie task count in `running_count`.

## Changes
- Introduced `@tasks_mutex : Mutex = Mutex.new` to synchronize all writes, deletions, and reads on `@running_tasks`.
- Wrapped subagent execution in `begin ... ensure @tasks_mutex.synchronize { @running_tasks.delete(task_id) } end` to guarantee task cleanup on unhandled exceptions.
- Added comprehensive unit specs in `spec/autobot/agent/subagent_spec.cr` testing active task tracking, exception cleanup, and concurrent fiber execution.

Closes #97.
Part of #92.

### DIFF
--- a/spec/autobot/agent/subagent_spec.cr
+++ b/spec/autobot/agent/subagent_spec.cr
@@ -23,6 +23,15 @@ private class BlockingSubagentManager < Autobot::Agent::SubagentManager
   ) : Nil
     @started.send(nil)
     @gate.receive
+  end
+
+  private def run_and_untrack(
+    task_id : String,
+    task : String,
+    label : String,
+    origin : Hash(String, String),
+  ) : Nil
+    super
   ensure
     @completed.send(nil)
   end
@@ -38,6 +47,15 @@ private class FailingSubagentManager < Autobot::Agent::SubagentManager
     origin : Hash(String, String),
   ) : Nil
     raise "Simulated background subagent failure"
+  end
+
+  private def run_and_untrack(
+    task_id : String,
+    task : String,
+    label : String,
+    origin : Hash(String, String),
+  ) : Nil
+    super
   ensure
     @completed.send(nil)
   end
@@ -46,7 +64,7 @@ end
 private class TrackingSubagentManager < Autobot::Agent::SubagentManager
   getter completed = Channel(String).new(100)
 
-  private def run_subagent(
+  private def run_and_untrack(
     task_id : String,
     task : String,
     label : String,
@@ -81,12 +99,6 @@ describe Autobot::Agent::SubagentManager do
     manager.gate.send(nil)
     manager.completed.receive
 
-    # Wait for outer ensure block to cleanly delete task
-    100.times do
-      break if manager.running_count == 0
-      sleep 1.millisecond
-    end
-
     manager.running_count.should eq(0)
   ensure
     FileUtils.rm_rf(tmp) if tmp
@@ -106,14 +118,8 @@ describe Autobot::Agent::SubagentManager do
     manager.running_count.should eq(0)
     manager.spawn("Failing task", label: "Failing Worker")
 
-    # Wait until background fiber finishes and raises
+    # Wait until background fiber finishes and cleans up
     manager.completed.receive
-
-    # Wait for outer ensure block to cleanly delete task
-    100.times do
-      break if manager.running_count == 0
-      sleep 1.millisecond
-    end
 
     manager.running_count.should eq(0)
   ensure
@@ -145,12 +151,6 @@ describe Autobot::Agent::SubagentManager do
 
     # Deterministically wait for all 25 subagent background fibers to complete
     concurrent_tasks.times { manager.completed.receive }
-
-    # Wait for outer ensure block to cleanly delete all tasks
-    100.times do
-      break if manager.running_count == 0
-      sleep 1.millisecond
-    end
 
     manager.running_count.should eq(0)
   ensure
--- a/spec/autobot/agent/subagent_spec.cr
+++ b/spec/autobot/agent/subagent_spec.cr
@@ -1,0 +1,159 @@
+require "../../spec_helper"
+
+private class MockSubagentProvider < Autobot::Providers::HttpProvider
+  def initialize
+    super(api_key: "test-key", model: "mock-model")
+  end
+
+  private def http_post(url : String, headers : HTTP::Headers, body : String) : HTTP::Client::Response
+    HTTP::Client::Response.new(200, body: %({"choices":[{"message":{"content":"Subagent finished"},"finish_reason":"stop"}],"usage":{"prompt_tokens":0,"completion_tokens":0,"total_tokens":0}}))
+  end
+end
+
+private class BlockingSubagentManager < Autobot::Agent::SubagentManager
+  getter started = Channel(Nil).new
+  getter gate = Channel(Nil).new
+  getter completed = Channel(Nil).new
+
+  private def run_subagent(
+    task_id : String,
+    task : String,
+    label : String,
+    origin : Hash(String, String),
+  ) : Nil
+    @started.send(nil)
+    @gate.receive
+  ensure
+    @completed.send(nil)
+  end
+end
+
+private class FailingSubagentManager < Autobot::Agent::SubagentManager
+  getter completed = Channel(Nil).new
+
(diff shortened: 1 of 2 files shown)


src/autobot/agent/subagent.cr
```diff
@@ -70,11 +71,18 @@ module Autobot
...
-          run_subagent(task_id, task, display_label, origin)
-          @running_tasks.delete(task_id)
+          begin
+            run_subagent(task_id, task, display_label, origin)
+          ensure
+            @tasks_mutex.synchronize do
+              @running_tasks.delete(task_id)
+            end
+          end
         end
```
veelenga, line 80: The fix itself looks right. One structural tweak makes the specs fully deterministic: extract this block into a named method so the deletion becomes an overridable seam (companion suggestion below on `running_count`).

```suggestion
        ::spawn do
          run_and_untrack(task_id, task, display_label, origin)
        end
```
Test subclasses can then override `run_and_untrack` and signal *after* `super` — i.e. strictly after the task is deleted — which removes every `sleep` polling loop from the specs. I ran the restructured specs 5x: 3/3 green, ameba clean.

src/autobot/agent/subagent.cr
```diff
@@ -83,7 +91,9 @@ module Autobot
 
       # Return the number of currently running subagents.
```
veelenga, line 86: Companion to the seam suggestion above — the extracted method:

```suggestion
      private def run_and_untrack(
        task_id : String,
        task : String,
        label : String,
        origin : Hash(String, String),
      ) : Nil
        run_subagent(task_id, task, label, origin)
      ensure
        @tasks_mutex.synchronize do
          @running_tasks.delete(task_id)
        end
      end

      # Return the number of currently running subagents.
```

spec/autobot/agent/subagent_spec.cr
```diff
@@ -0,0 +1,159 @@
...
+    task_id : String,
+    task : String,
+    label : String,
+    origin : Hash(String, String),
+  ) : Nil
+    @started.send(nil)
+    @gate.receive
+  ensure
+    @completed.send(nil)
+  end
```
veelenga, line 37: With the `run_and_untrack` seam in `SubagentManager`, signal completion from an override of it instead — `@completed.send` then runs strictly after the task deletion in `super`'s `ensure`, so the test below can assert `running_count == 0` immediately, no polling:

```suggestion
    @started.send(nil)
    @gate.receive
  end

  private def run_and_untrack(
    task_id : String,
    task : String,
    label : String,
    origin : Hash(String, String),
  ) : Nil
    super
  ensure
    @completed.send(nil)
  end
```

spec/autobot/agent/subagent_spec.cr
```diff
@@ -0,0 +1,159 @@
...
+  private def run_subagent(
+    task_id : String,
+    task : String,
+    label : String,
+    origin : Hash(String, String),
+  ) : Nil
+    raise "Simulated background subagent failure"
+  ensure
+    @completed.send(nil)
+  end
```
veelenga, line 61: Same pattern here — move the signal to a `run_and_untrack` override so it fires after cleanup:

```suggestion
    raise "Simulated background subagent failure"
  end

  private def run_and_untrack(
    task_id : String,
    task : String,
    label : String,
    origin : Hash(String, String),
  ) : Nil
    super
  ensure
    @completed.send(nil)
  end
```

spec/autobot/agent/subagent_spec.cr
```diff
@@ -0,0 +1,159 @@
...
+end
+
+private class TrackingSubagentManager < Autobot::Agent::SubagentManager
+  getter completed = Channel(String).new(100)
+
+  private def run_subagent(
```
veelenga: Same pattern — override `run_and_untrack` instead, so each `completed.receive` guarantees that task is already untracked:

```suggestion
  private def run_and_untrack(
```

spec/autobot/agent/subagent_spec.cr
```diff
@@ -0,0 +1,159 @@
...
+    manager.gate.send(nil)
+    manager.completed.receive
+
+    # Wait for outer ensure block to cleanly delete task
+    100.times do
+      break if manager.running_count == 0
+      sleep 1.millisecond
+    end
+
+    manager.running_count.should eq(0)
```
veelenga: With the signal moved after cleanup, `completed.receive` already guarantees the deletion happened — the polling loop can go:

```suggestion
    manager.running_count.should eq(0)
```

veelenga (commented):
Nice, the mutex + `ensure` fix is exactly right and I'll take it. One request on the specs: the three `sleep 1.millisecond` polling loops exist only because `@completed` is signalled from `run_subagent`, which runs *before* the spawn block's outer `ensure` deletes the task. Extracting that block into a `run_and_untrack` method (suggestions inline) lets the test subclasses signal after the deletion instead — all three loops disappear and the specs become deterministic. I verified the full set of suggestions locally: 3/3 specs pass, stable across 5 runs, ameba clean.

spec/autobot/agent/subagent_spec.cr
```diff
@@ -0,0 +1,159 @@
...
+    # Wait until background fiber finishes and raises
+    manager.completed.receive
+
+    # Wait for outer ensure block to cleanly delete task
+    100.times do
+      break if manager.running_count == 0
+      sleep 1.millisecond
+    end
+
+    manager.running_count.should eq(0)
```
veelenga:
```suggestion
    manager.running_count.should eq(0)
```

spec/autobot/agent/subagent_spec.cr
```diff
@@ -0,0 +1,159 @@
...
+    # Deterministically wait for all 25 subagent background fibers to complete
+    concurrent_tasks.times { manager.completed.receive }
+
+    # Wait for outer ensure block to cleanly delete all tasks
+    100.times do
+      break if manager.running_count == 0
+      sleep 1.millisecond
+    end
+
+    manager.running_count.should eq(0)
```
veelenga:
```suggestion
    manager.running_count.should eq(0)
```

renich:
Thanks for the excellent suggestion, Vitalii! I extracted `run_and_untrack` and updated the test subclasses to signal completion from overrides of it. All `sleep` polling loops are gone, and the specs are completely deterministic. Pushed!

veelenga (approved):

PR #103, opened 2026-08-30
fix(agent): propagate max_tokens and prevent empty response delivery

### LINKED ISSUE #106: Agent loop drops max_tokens/temperature and emits empty messages on blank LLM response
### Description
1. While `max_tokens` and `temperature` can be configured under `agents.defaults`, they are not forwarded through `ToolExecutor` to `Provider#chat`.
2. When an LLM returns an empty string or whitespace-only response (e.g., on token exhaustion or empty generations), the blank string is recorded in memory and dispatched to channels, triggering Telegram API 400 Bad Request errors (`Bad Request: message text is empty`).

### Expected Behavior
- Configured options (`max_tokens`, `temperature`) should be passed to `Provider#chat`.
- Empty or whitespace-only LLM responses should fall back to a default response string.
- `TelegramChannel#send_message` should ignore empty or whitespace-only messages.

renich:
## Summary
Propagates `max_tokens` and `temperature` from config through Gateway, Loop, SubagentManager, and ToolExecutor down to `Provider#chat`. In addition, prevents empty LLM responses from being emitted or recorded in session histories and adds empty outbound message protection in `TelegramChannel`.

### Root Cause
1. While `max_tokens` and `temperature` can be specified under `agents.defaults`, these options were not being forwarded through `ToolExecutor` to `Provider#chat`.
2. When an LLM returns an empty string or whitespace-only response (e.g. following certain tool invocations or token exhaustion), the empty string was saved to memory and dispatched to channels, leading to empty messages and Telegram API 400 Bad Request errors (`Bad Request: message text is empty`).

### Changes
- Propagate `max_tokens` and `temperature` in `ToolExecutor`, `Agent::Loop`, and `SubagentManager`.
- Use `String#presence || FALLBACK_RESPONSE` in `Agent::Loop` and `SubagentManager` so empty LLM responses default to a helpful fallback message rather than emitting blank strings.
- Guard `TelegramChannel#send_message` against empty or whitespace-only strings to avoid sending invalid payload to Telegram API.
- Add regression specs across `loop_spec.cr`, `tool_executor_spec.cr`, and `telegram_spec.cr`.

Closes #106

### DIFF
--- a/spec/autobot/agent/loop_spec.cr
+++ b/spec/autobot/agent/loop_spec.cr
@@ -2,12 +2,19 @@ require "../../spec_helper"
 
 # Mock provider that returns a simple text response (no tool calls).
 class MockProvider < Autobot::Providers::HttpProvider
-  def initialize
+  def initialize(@response_content : String = "Mock response", @responses : Array(String)? = nil)
     super(api_key: "test-key", model: "mock-model")
+    @response_index = 0
   end
 
   private def http_post(url : String, headers : HTTP::Headers, body : String) : HTTP::Client::Response
-    HTTP::Client::Response.new(200, body: %({"choices":[{"message":{"content":"Mock response"},"finish_reason":"stop"}],"usage":{"prompt_tokens":0,"completion_tokens":0,"total_tokens":0}}))
+    if responses = @responses
+      resp = responses[@response_index]? || responses.last
+      @response_index += 1
+      HTTP::Client::Response.new(200, body: resp)
+    else
+      HTTP::Client::Response.new(200, body: %({"choices":[{"message":{"content":#{@response_content.to_json}},"finish_reason":"stop"}],"usage":{"prompt_tokens":0,"completion_tokens":0,"total_tokens":0}}))
+    end
   end
 end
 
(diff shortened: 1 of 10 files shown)


src/autobot/cli/gateway.cr
```diff
@@ -159,7 +159,9 @@ module Autobot
           exec_deny_patterns: deny_patterns,
           exec_allow_patterns: allow_patterns,
           sandbox_config: sandbox_config,
-          rate_limiter: rate_limiter
+          rate_limiter: rate_limiter,
+          max_tokens: config.agents.try(&.defaults.try(&.max_tokens)),
```
veelenga: When the config has no `agents:` section this passes `nil` and the executor falls back to `Providers::DEFAULT_MAX_TOKENS` (4096). The documented default and the CLI `agent` command both use 8192 from `Config::AgentDefaults`. Please resolve the defaults here the same way `cli/agent.cr` does:

```crystal
agent_defaults = Config::AgentDefaults.new
defaults = config.agents.try(&.defaults)
max_tokens: defaults.try(&.max_tokens) || agent_defaults.max_tokens,
temperature: defaults.try(&.temperature) || agent_defaults.temperature
```

Then `Loop`, `SubagentManager` and `ToolExecutor` can take non-nil `Int32`/`Float64` and the `|| DEFAULT_*` fallbacks in the executor go away.

src/autobot/agent/tool_executor.cr
```diff
@@ -109,7 +111,9 @@ module Autobot::Agent
       response = @provider.chat(
         messages: messages,
         tools: tools.definitions(exclude: exclude_tools, compact: compact_tools),
-        model: @model
+        model: @model,
+        max_tokens: @max_tokens || Providers::DEFAULT_MAX_TOKENS,
```
veelenga: With the gateway resolving defaults (see comment there), make these plain `Int32`/`Float64` params and drop the `||` fallbacks. Two places deciding the default is how the 4096 vs 8192 mismatch happens.

src/autobot/agent/loop.cr
```diff
@@ -170,7 +178,7 @@ module Autobot::Agent
 
       @message_tool.try(&.clear_last_sent)
       result = @executor.execute(messages, @tools, session_key: session.key)
-      final_content = result.content || FALLBACK_RESPONSE
+      final_content = result.content.presence || FALLBACK_RESPONSE
```
veelenga: This changes what gets saved to the session in a common case. When the LLM replies via the `message` tool and then returns a blank final string, the outbound response is correctly suppressed below, but `FALLBACK_RESPONSE` ("I apologize, but I couldn't generate a response.") is now stored as the assistant turn. The next turn's history then claims the bot failed when it actually answered.

The cron path already handles this by preferring the message tool's last sent content. Please do the same here, e.g.:

```crystal
final_content = result.content.presence || @message_tool.try(&.last_sent_content) || FALLBACK_RESPONSE
```

and add a spec covering "message tool sent, LLM returned empty" that checks the saved history.

spec/autobot/channels/telegram_spec.cr
```diff
@@ -706,4 +706,18 @@ describe Autobot::Channels::TelegramChannel do
...
+      channel = build_channel
+      outbound = Autobot::Bus::OutboundMessage.new(
+        channel: "telegram",
+        chat_id: "123",
+        content: "   \n  "
+      )
+
+      # Should return immediately without raising or trying to send network requests
+      channel.send_message(outbound)
+    end
```
veelenga, line 848: This test has no assertion and passes with or without the guard: `api_request` rescues and logs every error, so without the guard it would silently make a real request to api.telegram.org from CI. The test helper already tracks clients, so please assert on that:

```crystal
channel.send_message(outbound)
channel.created_clients.should be_empty
```

Also drop the comment, the assertion says it.

spec/autobot/channels/telegram_spec.cr
```diff
@@ -603,7 +603,7 @@ describe Autobot::Channels::TelegramChannel do
       port = server.local_address.port
 
       spawn do
-        if socket = server.accept?
+        while socket = server.accept?
```
veelenga: Unrelated to this PR and not mentioned in the description. Please revert, or explain why it is needed.

spec/autobot/agent/loop_spec.cr
```diff
@@ -261,6 +261,37 @@ describe Autobot::Agent::Loop do
...
+      tools = Autobot::Tools::Registry.new
+      sessions = Autobot::Session::Manager.new(tmp)
+
+      loop_inst = TestableLoop.new(
+        bus: bus,
+        provider: provider,
+        workspace: tmp,
+        tools: tools,
+        sessions: sessions,
+        max_tokens: 16384,
```
veelenga: `max_tokens` and `temperature` are passed but never checked, so they are noise in a test about the empty fallback. Remove them, and match the sibling tests by passing `memory_window: 0` and `sandbox_config: "none"` so this test does not run sandbox detection.

veelenga (changes requested):
Thanks, the core fix is right: the gateway never passed `max_tokens`/`temperature` to the executor before, so every call used the provider default of 4096. Format, ameba and the full spec suite pass locally. A few things need to change before merge, see inline comments.

Also, please drop the unrelated `if` -> `while` change in the keep-alive spec unless there is a reason for it (scope boundary rule in CLAUDE.md).

Out of scope but worth a follow-up issue: `MemoryManager#summarize_to_memory` still calls the provider without `max_tokens`, so memory consolidation stays capped at 4096.

renich:
Thanks for the review and suggestions! I have updated the branch and addressed each item:

1. **Defaults Resolution in Gateway**:
   - Resolved defaults in `src/autobot/cli/gateway.cr` via `Config::AgentDefaults.new` matching `cli/agent.cr`.
   - Updated `Loop`, `SubagentManager`, and `ToolExecutor` to take plain `Int32` / `Float64` values with standard `Config::AgentDefaults` defaults, removing redundant `||` fallbacks in `ToolExecutor#call_llm`.

2. **Session History on Message Tool Turns**:
   - In `Agent::Loop#process_standard_message`, set `final_content = result.content.presence || @message_tool.try(&.last_sent_content) || FALLBACK_RESPONSE`.
   - When the LLM answers via the `message` tool and completes with an empty string, session history now accurately records the content sent via the tool rather than `FALLBACK_RESPONSE`.

3. **Specs Clean-up & New Coverage**:
   - Added a spec in `spec/autobot/agent/loop_spec.cr` verifying session history captures the message tool content on an empty LLM string.
   - Cleaned up the empty fallback test by removing unused parameter noise and passing `memory_window: 0`, `sandbox_config: "none"`.
   - In `spec/autobot/channels/telegram_spec.cr`, reverted the unrelated keep-alive spec change and added `channel.created_clients.should be_empty` to verify no request was attempted on empty messages.

4. **Rebased onto `main`**:
   - Rebased cleanly on top of upstream `main` (`v0.8.0`). All 1,121 specs and Ameba pass clean locally.

PR #104, opened 2026-08-30
fix(providers): send image blocks to Gemini and keep Telegram image documents for vision

### LINKED ISSUE #107: Inbound attachments not persisted to workspace and Gemini provider drops multimodal arrays
### Description
1. Inbound attachments (images/files) sent to channels are only held in memory as Base64 strings, leaving agent filesystem tools (`exec`, `read_file`, Python scripts) unable to access media on disk.
2. In `GeminiProvider`, `build_native_parts` only handles plain strings. Multimodal content arrays (`[{"type": "text"}, {"type": "image_url"}]`) generated by `Context::Builder` return nil, terminating requests on a `model` turn and causing HTTP 400 errors (`Requests ending with a model turn are not supported`).
3. Telegram document attachments with `image/*` MIME types are not recognized as photos and lack automatic binary downloading.

### Expected Behavior
- Inbound media should be decoded and persisted to `workspace/attachments/`.
- `GeminiProvider` should parse multimodal content blocks into native `inlineData` parts.
- Telegram documents with image MIME types should be downloaded and treated as photo attachments.

renich:
## Summary
Makes Gemini requests work when the user turn carries images, and lets Telegram documents that are really images reach the model as vision input.

### Root cause
1. `GeminiProvider#build_native_parts` only handled string content. When `Context::Builder` produced a multimodal array (`text` + `image_url` blocks) the user turn was dropped, so the request ended on a `model` turn and Gemini answered 400 `Requests ending with a model turn are not supported`.
2. `TelegramMedia#extract_document` never set `data` for documents, so images sent as uncompressed files were described to the model but not shown to it.

### Changes
- `GeminiProvider`: map `text` blocks to `text` parts and `image_url` data URIs to `inlineData` parts.
- `Provider`: share the base64 data URI parser between the HTTP and Gemini providers.
- `TelegramMedia`: keep the downloaded bytes as `data` when a document's MIME type is `image/*`.
- Specs for both paths in `gemini_provider_spec.cr` and `telegram_media_spec.cr`.

Closes #107

### DIFF
--- a/spec/autobot/channels/telegram_media_spec.cr
+++ b/spec/autobot/channels/telegram_media_spec.cr
@@ -170,6 +170,19 @@ describe Autobot::Channels::TelegramMedia do
       attachments.first.mime_type.should eq("application/pdf")
     end
 
+    it "keeps document bytes for vision when mime type is an image" do
+      media = build_media(bytes: "png-bytes".to_slice)
+      msg = message(%({"document": {"file_id": "d2", "file_name": "photo.png", "mime_type": "image/png"}}))
+
+      parts, attachments = media.extract(msg, typed_text: false)
+
+      parts.should eq(["[document: photo.png]"])
+      attachment = attachments.first
+      attachment.name.should eq("photo.png")
+      attachment.mime_type.should eq("image/png")
+      attachment.data.should eq(Base64.strict_encode("png-bytes"))
+    end
+
     it "returns nothing for a plain text message" do
       media = build_media
       parts, attachments = media.extract(message(%({"text": "hello"})), typed_text: true)
--- a/spec/autobot/providers/gemini_provider_spec.cr
+++ b/spec/autobot/providers/gemini_provider_spec.cr
@@ -39,4 +39,30 @@ describe Autobot::Providers::GeminiProvider do
     parts[1]["inlineData"]["mimeType"].as_s.should eq("image/png")
     parts[1]["inlineData"]["data"].as_s.should eq("iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNk+A8AAQUBAScY42YAAAAASUVORK5CYII=")
   end
+
+  it "drops image blocks whose url is not a base64 data URI" do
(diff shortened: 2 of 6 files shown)


veelenga:
Thanks for this, and sorry for the delay. Status after recent changes on main:

- **Inbound media persistence** landed in #111 in a slightly different shape: incoming files are saved under `<workspace>/inbox` (configurable via `media.inbox`), and `MediaAttachment` now carries `file_path`, `origin`, `transcript` and friends. Audio files, photos, documents and forwards are classified as content there. So this part of the PR is superseded, and the Telegram code it patches (`append_document_attachment`) no longer exists; media extraction moved to `src/autobot/channels/telegram_media.cr`.
- **Gemini multimodal parsing** (`append_multimodal_blocks` / `append_image_url_block`) is still needed and is not covered by anything on main. That part looks right and is self-contained.
- **Documents with an `image/*` MIME type treated as photos** is still a gap too. In the new extractor it would be a small change in `TelegramMedia#extract_document`: download the bytes as it already does, and set `data` when the MIME type starts with `image/`.

Would you be up for rebasing onto main, dropping the persistence part, and keeping the Gemini and document-image pieces? Both should merge cleanly. If you would rather not, say so and we will land those two pieces in a small follow-up with credit to you.

renich:
Thanks Vitalii! As proposed, I have rebased onto `main` and trimmed the PR down to just the two needed parts:

1. **Gemini Multimodal Parsing**:
   - Added `append_multimodal_blocks` and `append_image_url_block` in `src/autobot/providers/gemini_provider.cr` to map OpenAI-compatible `image_url` data URIs into native Gemini `inlineData` parts.
   - Added unit spec in `spec/autobot/providers/gemini_provider_spec.cr`.

2. **Telegram Document Images**:
   - In `TelegramMedia#extract_document` (`src/autobot/channels/telegram_media.cr`), set `data: Base64.strict_encode(bytes)` whenever the detected format MIME type starts with `image/`.
   - Added regression spec in `spec/autobot/channels/telegram_media_spec.cr`.

Dropped the superseded inbox persistence code. All checks and Ameba pass cleanly.

PR #67, opened 2026-06-10
feat(telegram): add voice, chat logs, and system metrics

renich:
### Description
This PR introduces Telegram plugins for handling voice messages, chat logging, and retrieving system metrics:
1. **TTS (Text-to-Speech) / Voice**: Added a TTS/voice plugin (`src/autobot/plugins/tts.cr`) that handles voice inputs and outputs.
2. **Chat Logs**: Added a chat logging plugin (`src/autobot/plugins/chat_log.cr`) to log Telegram conversations.
3. **System Metrics**: Added a system metrics plugin (`src/autobot/plugins/system_info.cr`) to retrieve host system diagnostics.
4. **Integration**: Integrated these plugins with the Telegram channel (`src/autobot/channels/telegram.cr`).

---

### AI Assistance Disclosure
This pull request was prepared with the assistance of Gemini AI. The user (Rénich) has directed, reviewed, and tested these changes, and takes full responsibility for the code.

Co-developed-by: Gemini AI <renich+gemini@woralelandia.com>
Signed-off-by: Rénich Bon Ćirić <renich@woralelandia.com>

PR #105, opened 2026-08-30
feat(telegram): support quotes, web_page articles, and rich message types

### LINKED ISSUE #108: Telegram channel ignores quotes, forward origins, and Instant View article previews
### Description
When users reply quoting a specific snippet, forward a channel post, or share web pages with Instant View previews, `TelegramChannel` only inspects top-level `text` or `caption`. As a result, quoted excerpts, forward attribution, Instant View article texts, polls, locations, and contacts are omitted from the agent's context.

### Expected Behavior
- `TelegramChannel` should extract and format quotes (`[Quoting: ...]`), forward origins (`[Forwarded from: ...]`), web page instant view articles, polls, locations, and contacts into the message text.

renich:
## Summary
Adds recursive parsing for Telegram quote blocks (`quote`), channel/user forward origin attributions (`forward_origin`), Instant View articles (`web_page`), stories, polls, locations, contacts, and nested rich message blocks (`pageBlock`).

### Root Cause
Inbound Telegram messages containing quoted text, forwarded channel posts, or shared web page previews had these fields omitted, causing the agent to miss essential conversation context or article summaries sent by users.

### Changes
- Implemented `build_content_and_media` in `TelegramChannel` to extract and format quotes (`[Quoting: ...]`), forward origins (`[Forwarded from: ...]`), link previews, Instant View articles, polls, locations, and contacts.
- Added safe type handling (`.try(&.as_h?)`) to prevent TypeCastError exceptions on malformed JSON payload fields.
- Added unit tests in `telegram_spec.cr` covering all quote, forward, poll, location, contact, and rich message scenarios.

Closes #108


src/autobot/channels/telegram.cr
```diff
@@ -633,19 +633,219 @@ module Autobot::Channels
...
+      if lpo = msg["link_preview_options"]?.try(&.as_h?)
+        if url = lpo["url"]?.try(&.as_s?)
+          wp_lines << "[Link: #{url}]" unless wp_lines.any?(&.includes?(url))
+        end
+      end
+
+      wp_lines.empty? ? nil : wp_lines.join("\n\n")
+    end
+
+    private def append_web_page_details(wp_lines : Array(String), web_page : JSON::Any) : Nil
```
veelenga: `web_page` is not a field of [`Message`](https://core.telegram.org/bots/api#message) in the Bot API — it's a TDLib/MTProto field (`messageText.web_page`). Bots never receive `title`, `description`, `site_name`, `author_name` or Instant View content. The only link-preview data a bot gets is [`LinkPreviewOptions`](https://core.telegram.org/bots/api#linkpreviewoptions) (`is_disabled`, `url`, `prefer_small_media`, `prefer_large_media`, `show_above_text`).

This method and its test exercise a payload the API cannot produce; suggest removing it.

src/autobot/channels/telegram.cr
```diff
@@ -633,19 +633,219 @@ module Autobot::Channels
...
+      quote = msg["quote"]?.try(&.as_h?)
+      return nil unless quote
+
+      quote_text = quote["text"]?.try(&.as_s?)
+      return nil unless quote_text
+
+      "[Quoting: #{quote_text.inspect}]"
+    end
+
+    private def extract_story_content(msg : JSON::Any) : String?
```
veelenga: [`Story`](https://core.telegram.org/bots/api#story) has exactly two fields: `chat` and `id`. There is no `caption` and no `story_id`, so lines 691 and 694 read fields that never exist. A faithful version is just:

```crystal
story_id = story["id"]?.try(&.as_i64?)
return nil unless story_id
```

Also, `[Story ID: ]` is emitted when the id is missing (line 696) — return `nil` instead.

src/autobot/channels/telegram.cr
```diff
@@ -633,19 +633,219 @@ module Autobot::Channels
...
+      when "channel"
+        chat_title = forward_origin["chat"]?.try(&.as_h?).try { |chat| chat["title"]?.try(&.as_s?) }
+        chat_title ? "[Forwarded from: #{chat_title}]" : nil
+      when "user"
+        sender_name = forward_origin["sender_user"]?.try(&.as_h?).try { |user| user["first_name"]?.try(&.as_s?) }
+        sender_name ? "[Forwarded from: #{sender_name}]" : nil
+      when "chat"
+        sender_chat = forward_origin["sender_chat"]?.try(&.as_h?).try { |chat| chat["title"]?.try(&.as_s?) }
+        sender_chat ? "[Forwarded from: #{sender_chat}]" : nil
+      when "story"
```
veelenga: There is no `story` message origin. [`MessageOrigin`](https://core.telegram.org/bots/api#messageorigin) is exactly one of `MessageOriginUser`, `MessageOriginHiddenUser`, `MessageOriginChat`, `MessageOriginChannel` (types `user`, `hidden_user`, `chat`, `channel`). Please drop this branch and its test.

For `channel`/`chat`, consider falling back to `chat.username` when `title` is absent, and [`author_signature`](https://core.telegram.org/bots/api#messageoriginchannel) is available if useful.

src/autobot/channels/telegram.cr
```diff
@@ -633,19 +633,219 @@ module Autobot::Channels
...
+      wp_lines << "Author: #{wp_author}" if wp_author
+      wp_lines << wp_desc if wp_desc
+      wp_lines << wp_url if wp_url
+
+      if nested_body = extract_rich_blocks_text(web_page)
+        wp_lines << nested_body unless wp_desc && nested_body.includes?(wp_desc)
+      end
+    end
+
+    private def append_raw_rich_blocks(wp_lines : Array(String), msg : JSON::Any) : Nil
```
veelenga: None of `message`, `page_blocks`, or `blocks` are fields of [`Message`](https://core.telegram.org/bots/api#message). Only `rich_message` is real, and its shape is [`RichMessage`](https://core.telegram.org/bots/api#richmessage) `{ blocks: Array(RichBlock), is_rtl }`.

Each [`RichBlock`](https://core.telegram.org/bots/api#richblock) has `type` (`paragraph`, `heading`, `preformatted`, `list`, `block_quotation`, …) and `text: RichText`. [`RichText`](https://core.telegram.org/bots/api#richtext) is a String, an Array of RichText, or `{type: "bold"|"italic"|…, text: RichText}`.

The `@type: "pageBlockParagraph"` / `richTextPlain` / `texts` structure in the test comes from TDLib's Instant View API, not the Bot API.

src/autobot/channels/telegram.cr
```diff
@@ -633,19 +633,219 @@ module Autobot::Channels
...
+      end
+    end
+
+    private def extract_rich_blocks_text(node : JSON::Any) : String?
+      collected = [] of String
+      collect_rich_node_texts(node, collected)
+      collected.empty? ? nil : collected.join(" ").strip
+    end
+
+    private def collect_rich_node_texts(node : JSON::Any, collected : Array(String)) : Nil
```
veelenga: Suggest replacing this generic tree walk with a renderer that follows the spec: [`RichBlock`](https://core.telegram.org/bots/api#richblock) by `type` and a small recursive [`RichText`](https://core.telegram.org/bots/api#richtext) (`String | Array | {type, text}`).

As written, the key blacklist below (`from`, `chat`, `sender_id`, `@type`) has to grow forever, and any `text` key inside `RichBlockButtons`, `RichBlockTable` cells, `RichTextButton` etc. gets concatenated into the prose. A `JSON::Serializable` struct per type gives compile-time field names that match the docs and removes the `.try(&.as_h?).try { … }` chains.

src/autobot/channels/telegram.cr
```diff
@@ -633,19 +633,219 @@ module Autobot::Channels
...
     end
 
+    private def extract_quote_content(msg : JSON::Any) : String?
+      quote = msg["quote"]?.try(&.as_h?)
+      return nil unless quote
+
+      quote_text = quote["text"]?.try(&.as_s?)
+      return nil unless quote_text
+
+      "[Quoting: #{quote_text.inspect}]"
```
veelenga: `String#inspect` escapes newlines to a literal `\n` and backslashes to `\\`, so a multi-line quote reaches the model mangled. Use plain quoting:

```crystal
"[Quoting: \"#{quote_text}\"]"
```

Also worth considering: the user is *replying to* the quote ([`TextQuote`](https://core.telegram.org/bots/api#textquote) — "quoted part of a message that is replied to by the given message"), so emitting `text` first and the quote after (or folding it into the existing `extract_reply_context`) reads more naturally.

src/autobot/channels/telegram.cr
```diff
@@ -633,19 +633,219 @@ module Autobot::Channels
...
+      end
+
+      if contact_text = extract_contact_content(msg)
+        content_parts << contact_text
+      end
+
       append_media_attachments(msg, content_parts, media_attachments)
 
-      content = content_parts.empty? ? "[empty message]" : content_parts.join("\n")
+      content = content_parts.empty? ? "[empty message]" : content_parts.join("\n\n")
```
veelenga: This changes the separator from `\n` to `\n\n` for **every** message (text + caption + media placeholders too), which is an unrelated global behaviour change. Please keep `\n` here or make the change in a separate commit with its own rationale.

src/autobot/channels/telegram.cr
```diff
@@ -633,19 +633,219 @@ module Autobot::Channels
...
+        chat_title ? "[Forwarded Story from: #{chat_title} (Story ID: #{story_id})]" : "[Forwarded Story (ID: #{story_id})]"
+      when "hidden_user"
+        sender_name = forward_origin["sender_user_name"]?.try(&.as_s?)
+        sender_name ? "[Forwarded from: #{sender_name}]" : nil
+      else
+        nil
+      end
+    end
+
+    private def extract_web_page_content(msg : JSON::Any) : String?
```
veelenga: Naming: `wp_lines`, `wp_hash`, `wp_site`, `lpo` → full words. And this method actually handles link previews *and* rich messages; it should be split (or, per the other comments, reduced to just `link_preview_options`).

src/autobot/channels/telegram.cr
```diff
@@ -633,19 +633,219 @@ module Autobot::Channels
...
+      story_id = story["id"]?.try(&.as_i64?) || story["story_id"]?.try(&.as_i64?)
+      chat = story["chat"]?.try(&.as_h?)
+      chat_title = chat ? (chat["title"]?.try(&.as_s?) || chat["username"]?.try(&.as_s?) || chat["first_name"]?.try(&.as_s?)) : nil
+      caption = story["caption"]?.try(&.as_s?)
+
+      story_header = chat_title ? "[Story from #{chat_title} (ID: #{story_id})]" : "[Story ID: #{story_id}]"
+      caption ? "#{story_header}\n#{caption}" : story_header
+    end
+
+    private def extract_forward_origin_content(msg : JSON::Any) : String?
```
veelenga, line 689: Minor: `build_content_and_media` is now nine near-identical `if x = extract_…` lines. A list of extractors + `compact_map` (or one `to_prompt` method per parsed type) keeps the orchestration flat as more message kinds get added.

spec/autobot/channels/telegram_spec.cr
```diff
@@ -706,4 +710,227 @@ describe Autobot::Channels::TelegramChannel do
...
+        "quote": {
+          "text": "Important article excerpt"
+        }
+      }))
+
+      content, _ = channel.test_build_content_and_media(msg)
+      content.should eq("[Quoting: \"Important article excerpt\"]\n\nWhat do you think about this?")
+    end
+
+    it "extracts web_page article preview when present without text" do
```
veelenga: Payload cannot occur: `web_page` is not a [`Message`](https://core.telegram.org/bots/api#message) field in the Bot API. Suggest only using fixtures copied from the docs.

spec/autobot/channels/telegram_spec.cr
```diff
@@ -706,4 +710,227 @@ describe Autobot::Channels::TelegramChannel do
...
+          "first_name": "Renich",
+          "last_name": "Bon"
+        }
+      }))
+
+      content, _ = channel.test_build_content_and_media(msg)
+      content.should eq("[Contact: Renich Bon (+521234567890)]")
+    end
+
+    it "extracts story content and caption" do
```
veelenga: `caption` is not a [`Story`](https://core.telegram.org/bots/api#story) field (only `chat`, `id`).

spec/autobot/channels/telegram_spec.cr
```diff
@@ -706,4 +710,227 @@ describe Autobot::Channels::TelegramChannel do
...
+          },
+          "caption": "Check out this update"
+        }
+      }))
+
+      content, _ = channel.test_build_content_and_media(msg)
+      content.should eq("[Story from Renich Channel (ID: 31)]\nCheck out this update")
+    end
+
+    it "extracts forwarded story origin" do
```
veelenga: `type: "story"` is not a valid [`MessageOrigin`](https://core.telegram.org/bots/api#messageorigin).

spec/autobot/channels/telegram_spec.cr
```diff
@@ -706,4 +710,227 @@ describe Autobot::Channels::TelegramChannel do
...
+          },
+          "story_id": 42
+        }
+      }))
+
+      content, _ = channel.test_build_content_and_media(msg)
+      content.should eq("[Forwarded Story from: News Feed (Story ID: 42)]\n\nLook at this story")
+    end
+
+    it "extracts nested rich message page blocks" do
```
veelenga: This is TDLib's Instant View schema (`@type`, `pageBlockParagraph`, `richTextPlain`, `texts`), not the Bot API's [`RichMessage`](https://core.telegram.org/bots/api#richmessage). A real fixture looks like:

```json
{"rich_message": {"blocks": [
  {"type": "paragraph", "text": ["This is ", {"type": "bold", "text": "paragraph one."}]}
]}}
```

veelenga (changes requested):
Thanks for the PR — the `quote`, `forward_origin` (user/hidden_user/chat/channel), `poll`, `location`, `contact` and `link_preview_options` handling matches the [Bot API](https://core.telegram.org/bots/api#message) and looks good.

The `web_page`, story caption/`story_id`, `forward_origin.type == "story"` and rich-message parsing are modelled on TDLib/MTProto types rather than the Bot API, so they parse payloads a bot can never receive. Details and doc links inline. The real [`RichMessage`](https://core.telegram.org/bots/api#richmessage) / [`RichBlock`](https://core.telegram.org/bots/api#richblock) / [`RichText`](https://core.telegram.org/bots/api#richtext) types are small and worth implementing directly instead of the generic tree walk.

renich:
Thanks for the detailed review and Bot API doc pointers! I have updated the branch with the following changes:

1. **Bot API Alignment**:
   - Dropped the MTProto/TDLib `web_page` extractor and `type: "story"` forward origin branch.
   - Refactored `rich_message` to follow the official Bot API `RichMessage` / `RichBlock` / `RichText` recursive schema (`paragraph`, `heading`, `block_quotation`, `preformatted`), removing the generic tree-walking heuristic.
   - Aligned `Story` parsing strictly to `chat` and `id` (`[Story from <title> (ID: <id>)]`), returning `nil` when `id` is absent.
   - Aligned `forward_origin` to `user`, `hidden_user`, `chat`, and `channel`, adding `chat.username` fallback when `title` is missing, and appending `author_signature` when present.
   - Reduced link preview handling to `link_preview_options` (`[Link: <url>]`).

2. **Formatting & String Ergonomics**:
   - Replaced `quote_text.inspect` with plain string interpolation `"[Quoting: \"#{quote_text}\"]"` to preserve newlines and backslashes without escape mangling.
   - Restored standard `\n` joining across message parts.
   - Replaced abbreviations with full-word identifiers.

3. **Specs**:
   - Replaced TDLib fixtures with realistic Bot API payloads matching the official documentation.

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,218 @@ module Autobot::Channels
...
+
+      rendered = render_rich_text(text_node).strip
+      return nil if rendered.empty?
+
+      case block_hash["type"]?.try(&.as_s?)
+      when "heading"
+        "### #{rendered}"
+      when "block_quotation"
+        "[Quoting: \"#{rendered}\"]"
+      when "preformatted"
```
veelenga: `"preformatted"` never matches a real payload: [RichBlockPreformatted](https://core.telegram.org/bots/api#richblockpreformatted)'s `type` is always the literal **`"pre"`**. A real `pre` block falls through to `else` and renders as plain text without code fences.

```suggestion
      when "pre"
```

It also has an optional `language` field you could put after the opening fence.

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,218 @@ module Autobot::Channels
...
+      text_node = block_hash["text"]?
+      return nil unless text_node
+
+      rendered = render_rich_text(text_node).strip
+      return nil if rendered.empty?
+
+      case block_hash["type"]?.try(&.as_s?)
+      when "heading"
+        "### #{rendered}"
+      when "block_quotation"
```
veelenga: Two problems with this branch:

1. [RichBlockBlockQuotation](https://core.telegram.org/bots/api#richblockblockquotation)'s `type` is always the literal **`"blockquote"`**, not `"block_quotation"`.
2. Its content lives in **`blocks: Array of RichBlock`** (plus optional `credit: RichText`) — there is **no `text` field**. So even with the type fixed, the `return nil unless text_node` guard at the top of `render_rich_block` drops every real block quotation.

Handling `blockquote` needs recursion into its nested `blocks` (e.g. render each nested block and join). Note that [RichBlockExpandableBlockQuotation](https://core.telegram.org/bots/api#richblockexpandableblockquotation) (`"expandable_blockquote"`) and [RichBlockPullQuotation](https://core.telegram.org/bots/api#richblockpullquotation) (`"pullquote"`) *do* have `text` — those already work via the `else` branch.

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,218 @@ module Autobot::Channels
...
+      return nil unless rich_message
+
+      blocks = rich_message["blocks"]?.try(&.as_a?)
+      return nil unless blocks
+
+      rendered_blocks = blocks.compact_map { |block| render_rich_block(block) }
+      rendered_blocks.empty? ? nil : rendered_blocks.join("\n")
+    end
+
+    private def render_rich_block(block : JSON::Any) : String?
```
veelenga, line 759: Minor: [RichBlockList](https://core.telegram.org/bots/api#richblocklist) (`"list"`) has `items: Array of RichBlockListItem` — each item has a `label` and nested `blocks` — and no `text`, so lists are silently dropped by the `text` guard. Lists commonly carry real article content; consider rendering `label` + nested blocks per item. Same pattern would cover `blockquote` (see below).

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,218 @@ module Autobot::Channels
...
+      return nil unless lat && lon
+
+      "[Location: #{lat}, #{lon}]"
+    end
+
+    private def extract_contact_content(msg : JSON::Any) : String?
+      contact = msg["contact"]?.try(&.as_h?)
+      return nil unless contact
+
+      phone = contact["phone_number"]?.try(&.as_s?) || ""
```
veelenga: Nit: unlike the other extractors, this one has no required-field guard — `"contact": {}` produces `[Contact:  ()]`. Suggest `return nil unless phone || !first_name.empty?`-style guard for consistency with `extract_location_content`/`extract_poll_content` (per [Contact](https://core.telegram.org/bots/api#contact), `phone_number` and `first_name` are required, so this is defensive only).

spec/autobot/channels/telegram_spec.cr
```diff
@@ -706,4 +710,238 @@ describe Autobot::Channels::TelegramChannel do
...
+                "This is ",
+                {"type": "bold", "text": "paragraph one."}
+              ]
+            },
+            {
+              "type": "heading",
+              "text": "Section Title"
+            },
+            {
+              "type": "block_quotation",
```
veelenga: This fixture can't occur: the API sends `"blockquote"` (with nested `blocks`, no `text` — [RichBlockBlockQuotation](https://core.telegram.org/bots/api#richblockblockquotation)) and `"pre"` ([RichBlockPreformatted](https://core.telegram.org/bots/api#richblockpreformatted)). As written, the test passes while the renderer misses both real payload shapes. Suggest:

```json
{"type": "blockquote", "blocks": [{"type": "paragraph", "text": "A quoted excerpt"}]},
{"type": "pre", "text": "code block", "language": "crystal"}
```

spec/autobot/channels/telegram_spec.cr
```diff
@@ -706,4 +710,238 @@ describe Autobot::Channels::TelegramChannel do
       server.close if server
     end
   end
+
+  describe "#build_content_and_media" do
+    it "extracts telegram quote text when message has quote" do
+      channel = build_channel
+      msg = JSON.parse(%({
```
veelenga: `./bin/ameba` flags all 17 of the new multi-line `%({...})` literals in this file with `Style/MultilineStringLiteral` (the base branch is at 0 for this file). CI has no lint job so checks stay green, but per repo conventions these should be `<<-JSON` heredocs.

veelenga (changes requested):
Re-reviewed after 2f765f69 — this is a big improvement. Resolved from the previous review: `web_page` parsing removed, `Story` now uses only `chat`+`id`, the invalid `story` forward origin is gone, the TDLib page-block tree walker is replaced with a spec-shaped `RichBlock` renderer, `inspect` mangling is fixed (with a multi-line spec), the `\n\n` separator change is reverted, naming is cleaned up, and lookups are safe (`as_h?`/`as_s?`).

renich:
Thanks for the thorough follow-up review! I have updated the branch with the following adjustments:

1. **`RichBlockPreformatted` Alignment**:
   - Switched the literal type check to `"pre"`.
   - Extracted the optional `language` field to format the markdown code fence as ````<language>` when present.

2. **`RichBlockBlockQuotation` Recursion**:
   - Switched the literal type check to `"blockquote"`.
   - Implemented recursion into nested `blocks: Array(RichBlock)`, rendering each nested block and quoting the combined result (`[Quoting: "..."]`).
   - Retained plain quote formatting for `expandable_blockquote` and `pullquote` which carry `text`.

3. **`RichBlockList` Support**:
   - Added support for `"list"` blocks by iterating through `items: Array(RichBlockListItem)`.
   - Formatted each list item with its `label` (or `-` fallback) and rendered child blocks.

4. **Defensive Guard on `extract_contact_content`**:
   - Added a defensive check to return `nil` if both `phone_number` and `first_name` are empty, preventing malformed `[Contact: ()]` output.

5. **Specs & Code Quality**:
   - Updated `spec/autobot/channels/telegram_spec.cr` fixtures with real Bot API schemas for `"blockquote"`, `"pre"` (with language), and `"list"`.
   - Added a test verifying empty `contact` payloads return `nil`.
   - All tests pass 100% green and Ameba static analysis passes with 0 failures.

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,266 @@ module Autobot::Channels
     end
 
     private def build_content_and_media(msg : JSON::Any) : {String, Array(Bus::MediaAttachment)}
```
veelenga: **Blocking — regression.** All four media appenders gate their placeholder *and* voice/audio transcription on `content_parts.empty?` (lines 908, 965, 977, 990). Before this PR only `text`/`caption` could fill the list, so that guard meant "no user text → describe the media". Now `extract_message_content_parts` pushes forward/story/quote/link/poll parts first, so:

```json
{"forward_origin": {"type": "user", "sender_user": {"id": 1, "is_bot": false, "first_name": "Bob"}},
 "voice": {"file_id": "v1", "file_unique_id": "u1", "duration": 3}}
```

renders as just `[Forwarded from: Bob]` — `transcribe_file` never runs and there is no `[voice message]` marker. Same for forwarded photos (`[photo]` lost), audio, and documents (`[document: report.pdf]` lost). Forwarding a voice note to the bot to get it transcribed is a common flow, so this is user-visible.

Suggested fix: keep `content_parts` as the *body* (text, caption, media) exactly as before, and prepend the attribution parts (forward, story) after the media appenders have run — the same pattern `prepend_reply_context` already uses. E.g. build `body_parts` from text/caption + media, then `content = (attribution_parts + body_parts + trailer_parts).join("\n")`.

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,266 @@ module Autobot::Channels
...
+      end
+
+      if contact_text = extract_contact_content(msg)
+        parts << contact_text
+      end
+
+      parts
+    end
+
+    private def extract_quote_content(msg : JSON::Any) : String?
```
veelenga: **Blocking — duplicate context.** Per the docs, `quote` is only set *"for replies that quote part of the original message"*, and `process_message` (line 565) already runs `prepend_reply_context(content, extract_reply_context(msg))`, which injects the **full** `reply_to_message.text`. So a quoted reply now produces:

```
[Replying to: "Long original … excerpt …"]

why?
[Quoting: "excerpt"]
```

The excerpt appears twice, the second time *after* the user's own text where it reads as something the user said, and it bypasses `REPLY_CONTEXT_MAX_LENGTH` (a 4 KB quote is emitted untruncated while the reply path caps at 500 chars).

Simplest correct version: drop `extract_quote_content` and have `extract_reply_context` prefer `msg["quote"]["text"]` over the full reply text when present — that is exactly what the quote *is* (the part of the replied-to message the user selected). That also folds it into the existing truncation.

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,266 @@ module Autobot::Channels
...
+      end
+
+      rendered_items.empty? ? nil : rendered_items.join("\n")
+    end
+
+    private def render_rich_text_block(block_hash : Hash(String, JSON::Any)) : String?
+      text_node = block_hash["text"]?
+      return nil unless text_node
+
+      rendered = render_rich_text(text_node).strip
```
veelenga: `.strip` runs before the `case`, so `pre` blocks lose the first line's leading indentation and any trailing newline, which corrupts relative indentation in code:

```json
{"type": "pre", "language": "python", "text": "    return x\n  y = 1"}
```
→
````
```python
return x
  y = 1
```
````

Only strip for the non-`pre` branches (or check emptiness with `rendered.blank?` and keep the raw text for `pre`).

File: src/autobot/channels/telegram.cr [hunk 23]
```diff
@@ -630,22 +630,266 @@ module Autobot::Channels
...
+
+      rendered_items = items.compact_map do |item|
+        item_hash = item.as_h?
+        next nil unless item_hash
+
+        label = item_hash["label"]?.try(&.as_s?)
+        item_blocks = item_hash["blocks"]?.try(&.as_a?)
+        next nil unless item_blocks
+
+        item_text = item_blocks.compact_map { |nested| render_rich_block(nested) }.join(" ").strip
```
veelenga: Item blocks are joined with `" "` while every other level joins with `"\n"`, so anything structural inside a list item breaks:

- a nested `list` → `1. Outer a. inner one\nb. inner two` (the inner `b.` becomes a top-level sibling);
- a `pre` inside an item → `1. Run: ```sh\nmake release\n```` with the fence opener mid-line, so it is no longer a code block.

Join with `"\n"` and indent continuation lines (or at least join with `"\n"`).

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,266 @@ module Autobot::Channels
...
+
+      "[Location: #{lat}, #{lon}]"
+    end
+
+    private def extract_contact_content(msg : JSON::Any) : String?
+      contact = msg["contact"]?.try(&.as_h?)
+      return nil unless contact
+
+      phone = contact["phone_number"]?.try(&.as_s?) || ""
+      first_name = contact["first_name"]?.try(&.as_s?) || ""
```
veelenga: The guard only rejects the both-empty case; the format string then assumes both are present. Since [`Contact`](https://core.telegram.org/bots/api#contact) requires both fields this is defensive-only, but the defensive path now produces malformed markers:

- `{"phone_number": "+15551234"}` → `[Contact:  (+15551234)]` (double space, empty name)
- `{"first_name": "Bob"}` → `[Contact: Bob ()]`

Simpler and correct for every combination:

```crystal
name = [contact["first_name"]?.try(&.as_s?), contact["last_name"]?.try(&.as_s?)].compact.join(" ").presence
phone = contact["phone_number"]?.try(&.as_s?).presence
return nil unless name || phone
phone ? "[Contact: #{name || "unknown"} (#{phone})]" : "[Contact: #{name}]"
```

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,266 @@ module Autobot::Channels
...
+          if opt_text = opt.as_h?.try(&.[]?("text")).try(&.as_s?)
+            poll_lines << "- #{opt_text}"
+          end
+        end
+      end
+
+      poll_lines.join("\n")
+    end
+
+    private def extract_location_content(msg : JSON::Any) : String?
```
veelenga: [`Message.venue`](https://core.telegram.org/bots/api#message): *"For backward compatibility, when this field is set, the location field will also be set."* So a shared venue currently renders as bare coordinates and its `title` / `address` are dropped:

```json
{"venue": {"location": {"latitude": 20.6597, "longitude": -103.3496}, "title": "Cafe Central", "address": "Av. Juarez 123"},
 "location": {"latitude": 20.6597, "longitude": -103.3496}}
```
→ `[Location: 20.6597, -103.3496]`

Check `venue` first and render `[Venue: Cafe Central, Av. Juarez 123 (20.6597, -103.3496)]`, falling back to `location` only when `venue` is absent. Minor: `Float64#to_s` switches to scientific notation below `1e-4` (`-0.00005` → `-5.0e-5`); `format("%.6f", …)` avoids that.

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,266 @@ module Autobot::Channels
...
+      else
+        "[Forwarded from: #{title}]"
+      end
+    end
+
+    private def extract_link_preview_options_content(msg : JSON::Any) : String?
+      link_preview = msg["link_preview_options"]?.try(&.as_h?)
+      return nil unless link_preview
+
+      if url = link_preview["url"]?.try(&.as_s?)
```
veelenga: Two small things: `is_disabled` is never read, so a sender who explicitly *disabled* the preview still gets `[Link: …]`; and `url` is normally already present in `text` (Telegram only sets `link_preview_options` when the options were changed), so the marker usually duplicates a URL the model already sees. Suggest `return nil if link_preview["is_disabled"]?.try(&.as_bool?)` and skip the marker when `text` already contains the URL.

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,266 @@ module Autobot::Channels
...
+    end
+
+    private def render_rich_blockquote(block_hash : Hash(String, JSON::Any)) : String?
+      nested_blocks = block_hash["blocks"]?.try(&.as_a?)
+      return nil unless nested_blocks
+
+      rendered = nested_blocks.compact_map { |nested| render_rich_block(nested) }.join("\n").strip
+      return nil if rendered.empty?
+
+      "[Quoting: \"#{rendered}\"]"
```
veelenga, line 780: Optional: `[Quoting: "…"]` is also the marker used for a reply [`TextQuote`](https://core.telegram.org/bots/api#textquote), i.e. *the user quoting part of the message they reply to*. Reusing it for a [`RichBlockBlockQuotation`](https://core.telegram.org/bots/api#richblockblockquotation) (plain article content) blurs the two meanings for the model. Since siblings already render as Markdown (`###`, fenced code, `1.` lists), a `> ` prefix per line is the consistent choice, with the optional `credit: RichText` on a trailing `— credit` line. Same for `expandable_blockquote` / `pullquote` at line 827.

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,266 @@ module Autobot::Channels
...
+      block_hash = block.as_h?
+      return nil unless block_hash
+
+      case block_hash["type"]?.try(&.as_s?)
+      when "blockquote"
+        render_rich_blockquote(block_hash)
+      when "list"
+        render_rich_list(block_hash)
+      else
+        render_rich_text_block(block_hash)
```
veelenga, line 769: Optional / follow-up: with the `text` guard in `render_rich_text_block`, documented blocks that carry real text are silently dropped: [`details`](https://core.telegram.org/bots/api#richblockdetails) (`summary` + `blocks`), [`table`](https://core.telegram.org/bots/api#richblocktable) (`cells`), `collage` / `slideshow` (`blocks`), and [`mathematical_expression`](https://core.telegram.org/bots/api#richblockmathematicalexpression) (`expression`). Verified: `[{"type":"details","summary":"FAQ","blocks":[{"type":"paragraph","text":"answer"}]}]` → `[empty message]`. Media-only blocks are fine to skip; `details` is cheap since it is `blockquote` plus a summary line.

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,266 @@ module Autobot::Channels
...
+    private def render_rich_text(node : JSON::Any) : String
+      if str = node.as_s?
+        str
+      elsif arr = node.as_a?
+        arr.map { |item| render_rich_text(item) }.join
+      elsif node_hash = node.as_h?
+        if text_node = node_hash["text"]?
+          render_rich_text(text_node)
+        else
+          ""
```
veelenga, line 836: Optional: two `RichText` leaves have no `text` and render as `""`: [`custom_emoji`](https://core.telegram.org/bots/api#richtextcustomemoji) (use `alternative_text`) and [`mathematical_expression`](https://core.telegram.org/bots/api#richtextmathematicalexpression) (use `expression`). Also worth considering: `url` / `email_address` / `phone_number` / `text_mention` entities are flattened to their label, so `["Read the ", {"type":"url","text":"release notes","url":"https://…"}]` → `Read the release notes` with the URL gone. Rendering `url` as `[text](url)` would match the Markdown convention used for the other blocks.

src/autobot/channels/telegram.cr
```diff
@@ -630,22 +630,266 @@ module Autobot::Channels
...
-        content_parts << caption
-      end
-
       append_media_attachments(msg, content_parts, media_attachments)
 
       content = content_parts.empty? ? "[empty message]" : content_parts.join("\n")
       {content, media_attachments}
     end
 
+    private def extract_message_content_parts(msg : JSON::Any) : Array(String)
```
veelenga: Still open from round 1 (non-blocking): ten copy-pasted `if x = extract_…(msg); parts << x; end` blocks. Cyclomatic complexity is 11 — over the "max complexity: 10" in `CLAUDE.md`, passing only because `.ameba.yml` raises the limit to 13. The equivalent single expression compiles to `Array(String)` and produces identical output:

```crystal
[
  extract_forward_origin_content(msg),
  extract_story_content(msg),
  msg["text"]?.try(&.as_s?),
  msg["caption"]?.try(&.as_s?),
  extract_quote_content(msg),
  extract_rich_message_content(msg),
  extract_link_preview_options_content(msg),
  extract_poll_content(msg),
  extract_location_content(msg),
  extract_contact_content(msg),
].compact
```

(If you take the blocking fix above, this naturally splits into attribution / body / trailer lists.)

spec/autobot/channels/telegram_spec.cr
```diff
@@ -706,4 +710,322 @@ describe Autobot::Channels::TelegramChannel do
...
+          }
+        }
+        JSON
+      )
+
+      content, _ = channel.test_build_content_and_media(msg)
+      content.should eq("[Link: https://github.com/crystal-lang/crystal]")
+    end
+
+    it "extracts forward_origin channel post attribution with signature and username fallback" do
```
veelenga: Spec structure (non-blocking): this `it` and the two below (`extracts user, chat, and hidden_user forward origins`, `extracts story content when id is present`) each run two or three independent Arrange-Act-Assert cycles, so a failure in the first scenario masks the rest. The repo's `autobot-test` skill asks for one AAA cycle per example — splitting them is mechanical. Also `returns nil for empty contact payload` (line 988) asserts `[empty message]`, not `nil`; rename to match.

veelenga (changes requested):
Re-reviewed 77ffb87 against the [Bot API](https://core.telegram.org/bots/api), type by type. Everything from the two previous rounds is resolved:

- `pre` / `blockquote` literals fixed; `blockquote` recurses into `blocks`; `pre` uses `language` for the fence; `list` renders `items[].label` + nested `blocks`.
- `extract_contact_content` has the required-field guard and a spec for `"contact": {}`.
- `quote`, `story`, `forward_origin` (all four `MessageOrigin` types, `author_signature`, `username` fallback), `link_preview_options`, `poll`, `location`, `contact` read only fields that exist on the documented types.
- Spec fixtures are `<<-JSON` heredocs matching the documented shapes. `./bin/ameba`, `crystal tool format --check`, `make release` and `crystal spec` (64 examples) all pass on the PR head.

Field names are now correct, but running real payload shapes through `build_content_and_media` surfaced two behavioural regressions that block merge, plus a few rendering bugs. Details inline.

**Blocking**

1. Forwarded / quoted / story media loses transcription and placeholders (see comment on `build_content_and_media`).
2. `quote` duplicates the reply context that `extract_reply_context` already injects (see comment on `extract_quote_content`).
3. Docs: `CLAUDE.md` rule 1 — "When making changes, ensure the docs/ also reflect it." `docs/telegram.md` → *Features* still lists only reply context, voice, photos and documents, and `docs/architecture.md:91` still says "Photos, voice, custom commands, allowlists". Please add the new inbound kinds (quotes, forwards, stories, polls, locations, contacts, rich messages) — earlier Telegram feature PRs (#50, #67) updated `docs/telegram.md` the same way.

**Should fix** (inline): `pre` indentation stripped, list items joined with a space, partial `contact` output, `venue` ignored, `is_disabled` ignored.

**Optional / non-blocking** (inline): rich `blockquote` reuses the reply-quote marker, blocks without `text` are dropped, link targets in `RichText` are flattened away, the still-open nit about the `if`-chain in `extract_message_content_parts`, and a couple of spec-structure notes.


**Consistency note on `as_s?` (line 557, outside the diff so not inline):** the PR describes switching to `as_s?` for "safe type handling", but `text` is still read with the raising `as_s` here (and in `extract_reply_context` line 629 and `addressed?` line 1394), so `"text": 123` raises before `build_content_and_media` ever runs, exactly as before. Either make all inbound reads `as_s?`-safe or leave the existing `as_s` in `build_content_and_media` too — half-and-half is confusing to the next reader.

renich:
Thanks for the thorough review and edge-case analysis! I have rebased the PR on latest `main` (`v0.8.0`) and addressed all points:

1. **Attribution vs. Body Separation (Media Placeholder & Transcription Regressions Fixed)**:
   - In `build_content_and_media`, structured extraction into `attribution_parts` (forwards, stories), `body_parts` (typed text + media), and `trailer_parts` (rich messages, link previews, polls, venues, contacts).
   - Because `media_extractor.extract(msg, !typed.nil?)` evaluates only whether user text was typed, forwarded voice notes, photos, audio, and documents now run transcriptions and inject placeholders properly before attribution is prepended (`[Forwarded from: Bob]\n[voice message]`).

2. **Quote Folded into Reply Context**:
   - Dropped `extract_quote_content` from the body.
   - In `extract_reply_context`, preferred `msg["quote"]?.try(&.[]?("text")).try(&.as_s?)` over full reply text. Quotes now naturally integrate into reply context, appear before the prompt, and respect `REPLY_CONTEXT_MAX_LENGTH`.

3. **Code Blocks & Lists Rendering**:
   - Preserved raw leading indentation and linebreaks in `pre` code blocks using `rendered.chomp` without stripping leading whitespace.
   - Joined list item blocks with `\n` to maintain structure for nested lists and fenced code blocks.

4. **Venue, Location, Link Preview, & Contact Enhancements**:
   - Checked `venue` before `location` to capture venue title, address, and coordinates, formatting latitude and longitude with `%.6f`.
   - Handled `is_disabled` in `link_preview_options` and skipped the link marker if the URL is already present in typed text.
   - Formatted contact entries with proper fallbacks (`[Contact: #{name || "unknown"} (#{phone})]` / `[Contact: #{name}]`).
   - Standardized on safe `as_s?` for inbound message parsing across `extract_reply_context` and `addressed?`.

5. **Docs & Specs**:
   - Updated `docs/telegram.md` and `docs/architecture.md` to document the new inbound features.
   - Refactored test examples in `spec/autobot/channels/telegram_spec.cr` into isolated, single-AAA assertions, including regression tests for forwarded voice notes and photos.
   - Passed all quality gates locally (1,137 specs green, Ameba clean).

veelenga (approved):
lgtm

PR #114, opened 2026-09-04
feat(tools): confirmation gate for mutating tools

veelenga:
## Overview

- [x] `tools.confirm` names tools that are held until the user types a four-character code in the chat, valid for five minutes, one pending call per session
- [x] the model receives the request as the tool result and is told to describe the call and ask for the code; typing it runs the held call and hands the outcome back to the model as the next turn
- [x] the code must be typed: a spoken message carries the transcription prefix and does not match, and attachment content never reaches the check
- [x] subagents cannot confirm, so gated tools are refused in background registries
- [x] MCP tools can be gated by name with the same `name` or `prefix*` syntax as `tools.enabled`
- [x] `autobot doctor` reports the gated tools and warns about an entry that matches no known tool
- [x] specs for the store, the registry gate, the loop round-trip, doctor and the schema; docs for configuration and security

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -105,6 +105,9 @@ tools:
     - list_dir
     - message
     - "ha_*"     # patterns ending in * match by prefix (skills, plugins, MCP tools)
+  confirm:       # Optional; these tools run only after the user types a code
+    - exec
+    - ha_call_service
   filesystem:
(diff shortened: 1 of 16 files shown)


veelenga:
Closing: decided against a typed confirmation step. Bots that can act on the world will rely on the tool allowlist, on receiving only the user's own words, and on not being handed attachments, rather than on a per-call code.

PR #136, opened 2026-09-15
fix(session): preserve canonical session keys with underscores in list_sessions

### LINKED ISSUE #135: fix(session): preserve canonical session keys containing underscores in list_sessions
### Description

When `Session::Manager#list_sessions` scans `.jsonl` session files in the workspace sessions directory, it reconstructs the session key from the filename using:

```crystal
Path[path].stem.gsub(\"_\", \":\")
```

Because `session_path` maps `:` to `_` when formatting filenames, `list_sessions` naively assumes that all underscores in the filename originated as colons. If a session key naturally contains underscores (such as `telegram:user_12345`, `slack:general_chat`, or a topic-scoped session identifier), `list_sessions` mutates genuine underscores into colons (`telegram:user:12345`).

Any caller that relies on `list_sessions` to inspect or resume sessions receives a corrupted key, leading to mismatched lookups or creating disjoint duplicate session files.

### Empirical Reproduction

```crystal
it \"preserves canonical session keys with underscores\" do
  manager = Autobot::Session::Manager.new(workspace: tmp)
  session = manager.get_or_create(\"telegram:user_12345\")
  session.add_message(\"user\", \"hello\")
  manager.save(session)

  sessions = manager.list_sessions
  sessions.first[\"key\"].should eq(\"telegram:user_12345\")
  # Expected: \"telegram:user_12345\"
  #      got: \"telegram:user:12345\"
end
```

### Root Cause & Proposed Fix

The first line of each session `.jsonl` file is a `Session::Metadata` record. Currently, `Metadata` stores timestamps and custom metadata, but omits the canonical session `key`.

1. Add `property key : String? = nil` to `Session::Metadata` and write `session.key` into the metadata record during `Manager#save`.
2. In `Manager#list_sessions`, read `data[\"key\"]?.try(&.as_s)` with fallback to `Path[path].stem.gsub(\"_\", \":\")` to preserve 100% backwards compatibility with legacy session files.

renich:
## Summary

Preserves canonical session keys that contain underscores when serializing and listing sessions.

### Root Cause

In `Session::Manager#list_sessions`, session keys were reconstructed via:
```crystal
Path[path].stem.gsub(\"_\", \":\")
```
Because `session_path` maps `:` to `_`, `list_sessions` assumed all underscores in filenames originated as colons. For session keys with genuine underscores (such as `telegram:user_12345`, `slack:general_chat`, or topic-scoped session keys), `list_sessions` mutated genuine underscores into colons (`telegram:user:12345`). Subsequent session lookups using these keys resulted in cache misses and fragmented or duplicate session files.

### Changes

- Added `key : String? = nil` to `Session::Metadata` and persisted `session.key` to the JSONL header on save.
- In `Session::Manager#list_sessions` and `Manager#load`, read `data[\"key\"]?.try(&.as_s?).presence` with fallback to filename stem substitution for backwards compatibility with legacy session files.
- Ensured `@cache` records both `session.key` and query `key` on load to prevent duplicate in-memory references.
- Added `require \"log\"` to `src/autobot/session/manager.cr`.
- Added unit and regression specs covering canonical keys with underscores, legacy files without `key`, empty/null key resilience, and cache synchronization.

Closes #135


veelenga:
Thanks for digging into this. `list_sessions` turned out to have no callers, so we're removing it in #141 instead of fixing it. The cache aliasing here would also bring deleted history back after `delete`. Closing in favour of #141.

PR #138, opened 2026-09-15
fix(tools): preserve exception messages in ToolRegistry for agent self-correction

### LINKED ISSUE #137: fix(tools): preserve exception messages in ToolRegistry for agent self-correction
### Problem

When a tool raises an unhandled exception during `tool.execute(params)`, `Autobot::Tools::Registry#execute` catches the exception in its `rescue ex : Exception` block, but completely discards `ex.message`:

```crystal
rescue ex : Exception
  error_msg = "Error executing #{name}"
  Log.error { error_msg }
  Log.error { ex.backtrace.join("\n") }
  ToolResult.error(error_msg)
end
```

The resulting `ToolResult` returned to the agent loop contains only the generic string `"Error executing <tool_name>"`.

### Root Cause & Impact

When tools like `BashTool`, `ExecTool`, `WebSearchTool`, or custom plugins encounter execution failures (e.g. `IO::Error`, `File::NotFoundError`, `JSON::ParseException`, or network connection drops), the LLM agent receives no detail about what failed or why. Without the underlying exception message:
1. The agent cannot self-correct (e.g. adjusting a query, fixing a malformed parameter, or switching to an alternative command).
2. The agent repeatedly invokes the failing tool with the same arguments or hallucinates failure reasons.
3. Debugging agent traces requires inspecting backend application logs rather than the tool interaction transcript.

### Proposed Solution

1. Preserve `ex.message` in `ToolRegistry#execute`:
   - If `ex.message` is present, format as `"Error executing #{name}: #{ex.message}"`.
   - If `ex.message` is absent or empty, fall back to `"Error executing #{name}"`.
2. Ensure credential redaction (`LogSanitizer.redact_credentials`) is applied to the error message so sensitive tokens/keys potentially present in exception messages are redacted before returning to the agent.
3. Add regression specs verifying exception message preservation, empty message fallback, and credential redaction on errors.

renich:
### Summary

Closes #137

When a tool raised an unhandled exception during `tool.execute(params)`, `Autobot::Tools::Registry#execute` previously logged the backtrace but discarded `ex.message`, returning only a generic `ToolResult.error("Error executing #{name}")`.

This starved the ReAct agent loop of failure feedback. When tools encounter network drops, missing files, malformed JSON, or invalid arguments, the LLM was unable to self-correct and either hallucinated failure causes or repeatedly retried identical calls.

### Changes

1. **Preserve Underlying Exception Message**:
   - Extract and strip `ex.message` via `ex.message.try(&.strip.presence)`.
   - If present, format as `"Error executing #{name}: #{details}"`.
   - If absent or empty, fall back to `ex.class.name` (for subclasses like `Socket::ConnectError` or custom errors) or `"Error executing #{name}"` for base `Exception`.
2. **Token Budget & Context Protection**:
   - Bound error details to `MAX_ERROR_LENGTH` (4,000 chars) with an explicit `"... (truncated)"` suffix to protect model prompt tokens from oversized error payloads (e.g. huge command stderr or HTML error responses).
3. **Credential Redaction Parity**:
   - Pass the error message through `LogSanitizer.redact_credentials` prior to logging and returning `ToolResult.error`, ensuring API keys, auth tokens, or passwords embedded in exception messages are redacted before reaching the model.
4. **Exception Safety & Method Extraction**:
   - Safely log backtraces with `if bt = ex.backtrace?` avoiding unsafe `.not_nil!` assertions.
   - Extracted `handle_execution_error` helper to keep `Registry#execute` cyclomatic complexity clean (<= 10).
5. **Specs**:
   - Added comprehensive regression tests in `spec/autobot/tools/registry_spec.cr` verifying message preservation, empty message fallback, whitespace-only message fallback, whitespace stripping, subclass name fallback, credential redaction, and oversized message truncation.

### Verification

- `crystal tool format --check`: 0 diffs.
- `./bin/ameba src/ spec/`: 157 files inspected, 0 failures.
- `crystal build src/main.cr -o /dev/null --release --no-debug --single-module`: Clean release build.
- `crystal spec`: 1,192 examples, 0 failures, 0 errors.


veelenga:
Thanks for raising this. I went with a smaller version in #142

PR #147, opened 2026-09-17
fix(agent): support CRLF line endings in skill frontmatter parser

### LINKED ISSUE #146: SkillsLoader fails to parse frontmatter on files with CRLF line endings
### Problem

When `SKILL.md` files use CRLF (`\r\n`) line endings (such as when created or edited in Windows environments or checked out with default git line ending conversions), `SkillsLoader.parse_frontmatter` fails to match the frontmatter block. As a result:
- Skill description, parameters, tool bindings, and requirements are lost (returning `SkillMetadata.new` with `nil` fields).
- `SkillsLoader#load_skill` fails to strip the frontmatter block via `strip_frontmatter`, leaving the raw YAML header prepended to the skill instructions provided to the model.

### Root Cause

In `src/autobot/agent/skills.cr`:
1. `SkillsLoader.parse_frontmatter` uses `content.match(/\A---\n(.*?)\n---/m)`, which expects Unix `\n` directly following `---`. On files with CRLF, the delimiter is `---\r\n`, causing the match to fail.
2. In `strip_frontmatter`, the regex `/\A---\n.*?\n---\n/m` similarly requires LF line endings, causing `strip_frontmatter` to return the entire content with the YAML header unstripped.
3. In `parse_frontmatter`, splitting with `match[1].split("\n")` leaves carriage return characters (`\r`) on line endings when not chomped.

### Reproduction

```crystal
content = "---\r\nname: deploy\r\ndescription: Deploy app\r\nalways: true\r\n---\r\n# Deploy Guide\r\n"
meta = Autobot::Agent::SkillsLoader.parse_frontmatter(content)
meta.description # => nil (expected "Deploy app")
```

### Proposed Solution

Support optional `\r` in the regex delimiters:
- In `parse_frontmatter`: `/\A---\r?\n(.*?)\r?\n---/m` and split with `.lines` (or chomp line breaks).
- In `strip_frontmatter`: `/\A---\r?\n.*?\r?\n---\r?\n?/m`.

renich:
### Summary

Support CRLF (`\r\n`) line endings alongside LF (`\n`) in `SkillsLoader.parse_frontmatter` and `SkillsLoader#strip_frontmatter`.

### Problem & Root Cause

When `SKILL.md` files use Windows/CRLF line endings, `SkillsLoader.parse_frontmatter` fails to match the frontmatter block because the delimiter regex expected `\A---\n`. Consequently:
1. All frontmatter metadata (`description`, `tool`, `params`, `requires`) was ignored, returning empty metadata.
2. `strip_frontmatter` failed to strip the frontmatter block, causing the raw YAML frontmatter to be prepended to the prompt context.

### Solution

1. Updated the frontmatter opening delimiter to match optional `\r` (`/\A---\r?\n.../m`).
2. Updated closing delimiters to `---(?:\r?\n|\z)` to cleanly match CRLF and EOF boundaries without premature matches.
3. Switched line splitting from `match[1].split("\n")` to `match[1].lines`, which strips carriage returns (`chomp: true` by default in Crystal) while keeping line structure intact.
4. Added unit and integration regression specs in `spec/autobot/agent/skills_spec.cr`.

Closes #146


veelenga (approved):
lgtm

PR #149, opened 2026-09-17
fix(sqlite): use SQL string literal escaping for recorded migration versions

### LINKED ISSUE #148: SQLite plugin fails to record migrations when filename contains single quotes
### Problem

In the built-in `sqlite` plugin, when a migration filename contains a single quote (e.g. `001_create_user's_profile.sql`), applying the migration executes the SQL file successfully, but recording the applied version in `schema_migrations` fails with an SQLite syntax error:

```text
Error: in prepare, unrecognized token: "\"
  INSERT INTO schema_migrations (version) VALUES ('001_create_user'\''s_profile.sql');
                                      error here ---^
```

Because recording fails after the schema modification has already executed, the database is left with the migration applied but unrecorded in `schema_migrations`. On subsequent `migrate` runs, Autobot tries to re-apply the migration, failing because tables/indices already exist.

### Root Cause

In `src/autobot/plugins/builtin/sqlite.cr:236`:
```crystal
record_sql = "INSERT INTO schema_migrations (version) VALUES (#{shell_escape(file)});"
record_result = @executor.exec(
  "sqlite3 -safe #{shell_escape(path)} #{shell_escape(record_sql)}",
  timeout: INIT_TIMEOUT
)
```

`shell_escape` formats the filename according to POSIX shell rules (`'\''`). However, `file` is an SQL string literal inside `record_sql`, which is then itself escaped for the shell on line 238 via `#{shell_escape(record_sql)}`.

Because SQL does not recognize backslash escaping in string literals, `sqlite3` rejects the syntax. In standard SQL, single quotes inside literals are escaped by doubling them (`''`).

### Reproduction

```crystal
tmp = TestHelper.tmp_dir
Dir.cd(tmp) do
  Dir.mkdir_p("data/migrations/test")
  File.write("data/migrations/test/001_create_user's_profile.sql", "CREATE TABLE profiles (id INTEGER PRIMARY KEY);")
  tool = Autobot::Plugins::Builtin::SQLiteTool.new(Autobot::Tools::SandboxExecutor.new(nil))
  result = tool.execute({"action" => JSON::Any.new("migrate"), "db" => JSON::Any.new("test")})
  result.success? # => false (STDERR: unrecognized token: "\")
end
```

### Proposed Solution

Use standard SQL string literal escaping for the migration filename when constructing `record_sql`:
```crystal
record_sql = "INSERT INTO schema_migrations (version) VALUES ('#{sql_escape(file)}');"
```
where `sql_escape` escapes single quotes as `''`.

renich:
### Summary

When recording an applied migration version into `schema_migrations`, `apply_single_migration` formatted the SQL statement using `shell_escape(file)`:
```crystal
record_sql = "INSERT INTO schema_migrations (version) VALUES (#{shell_escape(file)});"
```
and then executed it via `@executor.exec("sqlite3 -safe #{shell_escape(path)} #{shell_escape(record_sql)}")`.

Because `shell_escape` formats arguments for a POSIX shell using `'\''`, embedding it inside an SQL string literal causes SQLite to receive backslash escapes rather than SQL literal quotes (`''`). When a migration filename contains a single quote (e.g. `001_create_user's_profile.sql`), SQLite fails with a syntax error:
```
Error: in prepare, unrecognized token: "\"
```
leaving the migration applied on disk but failing to record in `schema_migrations`.

### Changes

- Add private `sql_quote(val : String) : String` helper to `SQLiteTool` escaping single quotes as `''`: `'#{val.gsub("'", "''")}'`.
- Use `sql_quote(file)` when constructing `record_sql` prior to outer shell argument escaping.
- Add regression spec in `spec/autobot/plugins/builtin/sqlite_spec.cr` verifying migrations containing single quotes in filenames apply and record cleanly.

Closes #148


veelenga:
Thanks

PR #152, opened 2026-09-18
fix(sandbox): use Process::Status[] and return timeout status on process timeout

### LINKED ISSUE #151: Sandbox timeout fallback uses Process::Status.new causing RuntimeError on exit_code
### Problem Description
In `src/autobot/tools/sandbox.cr`, when a process execution times out and the termination rescue block triggers in `wait_for_process`:
```crystal
rescue
  Process::Status.new(TIMEOUT_EXIT_CODE)
end
```
`Process::Status.new(exit_status : Int32)` is deprecated in Crystal stdlib and assigns the raw OS system exit status. On Unix, raw exit status `124` is treated as a termination signal (`Signal[124]`) rather than an exit code (which is encoded in bits 8-15).

When callers (such as `SandboxExecutor#format_exec_output` or `SandboxExecutor#require_success`) inspect the timed-out process via `status.exit_code`, an unhandled exception is raised:
```
RuntimeError: Abnormal exit has no exit code
```

### Empirical Reproduction
```crystal
status = Process::Status.new(124)
status.exit_code # => Unhandled exception: Abnormal exit has no exit code (RuntimeError)
```

### Expected Behavior
`wait_for_process` should use `Process::Status[TIMEOUT_EXIT_CODE]`, which is the portable, non-deprecated constructor. On Unix systems, this properly encodes the exit status (`code << 8`), setting `normal_exit?` to `true` and allowing `status.exit_code` to cleanly return `124`.

```crystal
status = Process::Status[124]
status.exit_code # => 124
status.normal_exit? # => true
status.success? # => false
```

renich:
In `src/autobot/tools/sandbox.cr`, `Process::Status.new(exit_status : Int32)` is deprecated in Crystal stdlib and assigns the raw OS system exit status. On Unix, raw exit status `124` is interpreted as a termination signal (`Signal[124]`) rather than an exit code, causing callers inspecting `status.exit_code` to raise `RuntimeError: Abnormal exit has no exit code`.

Additionally, `wait_for_process` previously returned `status = process.wait` when process termination did not raise, returning a signal exit status instead of timeout status.

This PR switches to `Process::Status[TIMEOUT_EXIT_CODE]` and returns it unconditionally after process reaping on timeout.

Closes #151


src/autobot/tools/sandbox.cr
```diff
@@ -359,11 +359,11 @@ module Autobot
...
             sleep SIGNAL_GRACE_PERIOD
             process.signal(Signal::KILL) unless process.terminated?
-            status = process.wait
-            status
+            process.wait
           rescue
-            Process::Status.new(TIMEOUT_EXIT_CODE)
+            # Process already terminated or reaped
           end
+          Process::Status[TIMEOUT_EXIT_CODE]
```
veelenga: this seems to be a breaking change for Crystal 1.20

renich: Good catch. `Process::Status.[]` was indeed added in Crystal 1.21.0, so it fails compilation on Crystal <= 1.20.

I have updated the implementation with a version-gated helper:

```crystal
private def self.timeout_status : Process::Status
  {% if compare_versions(Crystal::VERSION, "1.21.0") >= 0 %}
    Process::Status[TIMEOUT_EXIT_CODE]
  {% else %}
    Process::Status.new({% if flag?(:unix) %} TIMEOUT_EXIT_CODE << 8 {% else %} TIMEOUT_EXIT_CODE {% end %})
  {% end %}
end
```

On Crystal >= 1.21.0, this uses the non-deprecated `Process::Status[]` constructor. On Crystal <= 1.20, it falls back to `Process::Status.new`, shifting by 8 on Unix so `normal_exit?` is true and `status.exit_code` returns 124 without raising an abnormal exit error.

PR #156, opened 2026-09-19
fix(tools): guard against empty old_text in EditFileTool

### LINKED ISSUE #155: EditFileTool enters infinite loop when old_text is empty
### Problem Description
In `src/autobot/tools/filesystem.cr`, when `EditFileTool` receives an empty string for `old_text`, it enters an infinite loop, spinning the fiber at 100% CPU.

### Root Cause
1. In `validate_and_replace`:
   `content.includes?(old_text)` evaluates to `true` because any string includes `""`.
2. Execution proceeds to `count_occurrences(content, "")`:
   ```crystal
   private def count_occurrences(haystack : String, needle : String) : Int32
     count = 0
     index = 0
     while pos = haystack.index(needle, index)
       count += 1
       index = pos + needle.size
     end
     count
   end
   ```
3. On the initial iteration, `haystack.index("", 0)` matches at position `0`.
4. `index = pos + needle.size` computes `0 + 0 = 0`.
5. Because `index` never advances, the loop condition continuously matches at `0`, freezing the fiber permanently.

### Expected Behavior
When `old_text` is empty, `EditFileTool` should return `ToolResult.error("old_text cannot be empty")` so the calling agent can self-correct. Additionally, `count_occurrences` should defensively return 0 if `needle.empty?` to prevent zero-width substring loops.

renich:
When `EditFileTool` receives an empty `old_text`, `validate_and_replace` bypassed the existence check (`"any string".includes?("")` returns `true` in Crystal) and invoked `count_occurrences`.

Inside `count_occurrences`, `haystack.index("", 0)` matched at index 0 and advanced by `needle.size` (0), causing `index` to remain 0 and locking up the executing fiber in an infinite loop at 100% CPU.

This PR:
1. Validates that `old_text` is non-empty in `validate_and_replace`, returning `ToolResult.error("old_text cannot be empty")` so calling agents receive actionable feedback to self-correct.
2. Defensively guards `count_occurrences` against empty search needles (`return 0 if needle.empty?`) to maintain an algorithmic $O(N)$ progress invariant.
3. Adds unit test coverage in `spec/autobot/tools/filesystem_spec.cr` verifying that empty `old_text` returns an error rather than hanging.

Closes #155


veelenga:
Verified the bug on `main` (`876bf81`) and the fix on this branch. One correction to the description: the loop is not infinite.

`count` is an `Int32`, so after about 2.1 billion iterations `count += 1` raises `OverflowError`. `Registry#execute` rescues it and the model gets `Error executing edit_file: Arithmetic overflow`.

Measured with the new spec applied to `main`:

| Build | Result |
|---|---|
| `main`, debug | 100% CPU for 22s, then `OverflowError` |
| `main`, release (minimal copy of `count_occurrences`) | 3.9s of CPU, then `OverflowError` |
| this branch | spec file passes in 11ms, 13 examples, 0 failures |

So the real impact is a multi-second freeze of the whole process (the loop never yields, so polling and cron stall too), followed by an error the model cannot act on and may retry. Still a real bug, and the guard fixes it.

renich:
Thank you for the empirical benchmark and precise breakdown, @veelenga! That's a great catch on the `Int32` `OverflowError` ceiling and the cooperative fiber stall impact. I'm glad the guard resolved it cleanly.

PR #158, opened 2026-09-21
fix(cron): enforce timeout and bound output buffers in direct exec

### LINKED ISSUE #157: Cron direct exec lacks execution timeout and output buffer bounding
### Problem Description
In `src/autobot/cron/service.cr`, when executing scheduled cron jobs with `kind: exec`:
- In sandboxed mode (`exec_command_sandboxed`), commands run via `Tools::Sandbox.exec`, which enforces `EXEC_TIMEOUT` (30s) and bounds output capture to prevent runaway buffer growth.
- In direct mode (`exec_command_direct`, when `sandbox: "none"`), commands run via raw `Process.run`:
  ```crystal
  output = IO::Memory.new
  error = IO::Memory.new
  status = Process.run(
    "sh", {"-c", command},
    output: output,
    error: error,
    env: env,
  )
  ```

### Root Cause
1. `Process.run` in Crystal stdlib does not accept a timeout; it blocks synchronously until the child process terminates. If a direct cron job hangs (e.g. hung network connection, interactive prompt, deadlocked script, or long sleep), the cron runner fiber hangs indefinitely, blocking all subsequent cron executions.
2. `output` and `error` are unbounded `IO::Memory` buffers. If a direct command emits large volumes of output, heap memory grows without bound until the process crashes or exhausts system RAM.

### Expected Behavior
Direct cron execution should mirror sandboxed execution by:
1. Enforcing an execution timeout (defaulting to `EXEC_TIMEOUT = 30.seconds`).
2. Bounding output buffers using `Tools::Sandbox.capture_command`.
3. Terminating/killing hung processes cleanly if the timeout expires.

renich:
Direct cron execution (`exec_command_direct`, when `sandbox: "none"`) previously ran commands via raw `Process.run` without an execution timeout. If a direct command hung or entered an indefinite wait, the cron runner fiber blocked indefinitely.

### Changes

- Extended `Tools::CommandRunner.run` with `env: env` and optional `max_output_size : Int32? = DEFAULT_MAX_OUTPUT`.
- Ran direct cron execution through `CommandRunner.run` with `max_output_size: nil`, allowing existing `truncate_output` logic to cap stored output at 32 KB without redundant limits.
- Extracted `handle_exec_result` in `Cron::Service` to unify outcome handling across sandboxed and direct execution paths, reporting timeouts natively as explicit timeout events (`command timed out after X seconds`).
- Added `cron.exec_timeout` (defaulting to 30s) to `CronConfig` and CLI gateway initialization.
- Added unit specs in `spec/autobot/cron/service_spec.cr`, `spec/autobot/tools/command_runner_spec.cr`, and `spec/autobot/config/schema_spec.cr`.
- Updated `docs/cron.md` and `docs/configuration.md` documenting the timeout and configuration.

Closes #157


src/autobot/cron/service.cr
```diff
@@ -479,20 +484,21 @@ module Autobot
...
+          timeout: @exec_timeout.total_seconds.to_i,
+          max_output_size: MAX_CAPTURED_OUTPUT_BYTES,
           env: env,
         )
 
         unless status.success?
-          raise "command exited with #{status.exit_code}: #{error.to_s.strip}"
+          exit_code = status.exit_code? || Tools::Sandbox::TIMEOUT_EXIT_CODE
+          err = stderr.strip
+          raise err.empty? ? "command exited with #{exit_code}" : "command exited with #{exit_code}: #{err}"
```
veelenga: `exit_code? || TIMEOUT_EXIT_CODE` turns both a timeout and a signal kill into `command exited with 124`, an exit code the command never returned. A user reading the job error cannot tell a timeout from a real `exit 124` or from an OOM kill.

Better to say what happened: `command timed out after N seconds` for a timeout, and the signal name when the command was killed (`Process::Status#to_s` gives `KILL`). `CommandRunner::Result` on `main` has `timed_out?` and `status` for exactly this.

src/autobot/cron/service.cr
```diff
@@ -454,11 +456,14 @@ module Autobot
...
-          timeout: EXEC_TIMEOUT.total_seconds.to_i,
+          timeout: @exec_timeout.total_seconds.to_i,
+          max_output_size: MAX_CAPTURED_OUTPUT_BYTES,
         )
 
         unless status.success?
-          raise "command exited with #{status.exit_code}: #{stderr.strip}"
+          exit_code = status.exit_code? || Tools::Sandbox::TIMEOUT_EXIT_CODE
+          err = stderr.strip
+          raise err.empty? ? "command exited with #{exit_code}" : "command exited with #{exit_code}: #{err}"
```
veelenga: This block is now repeated word for word in `exec_command_direct`. Both paths end up with the same kind of result, so one private method that takes the result and either returns `stdout.strip` or raises would keep the two paths from drifting apart.

src/autobot/cron/service.cr
```diff
@@ -18,8 +18,9 @@ module Autobot
 
       # Cap stored exec output so it can be safely re-exported as PREV_OUTPUT on
       # the next run without tripping E2BIG (argument list too long).
-      MAX_STORED_OUTPUT_BYTES  = 32_768
-      OUTPUT_TRUNCATION_NOTICE = "\n...[Output truncated due to size limit]"
+      MAX_STORED_OUTPUT_BYTES   =  32_768
+      MAX_CAPTURED_OUTPUT_BYTES = 100_000
```
veelenga: I do not think this constant is needed. `run_exec_job` cuts the output to `MAX_STORED_OUTPUT_BYTES` (32 KB) right after the command returns, so capturing up to 100 KB only to drop two thirds of it buys nothing. It also raises the sandboxed capture limit from 10 KB to 100 KB as a side effect, which is not part of this fix. Bounding the direct path with the existing limit is enough.

src/autobot/cron/service.cr
```diff
@@ -37,6 +38,7 @@ module Autobot
         @on_exec : ExecCallback? = nil,
         @workspace : Path? = nil,
         @sandbox_config : String = "none",
+        @exec_timeout : Time::Span = EXEC_TIMEOUT,
```
veelenga, line 40: Injecting the timeout is a reasonable way to test this without a 30s wait. As it stands only the spec can set it, though. Since the limit is new for `sandbox: none`, this would be the natural value to read from config, so users with longer unsandboxed commands have a way out.

spec/autobot/cron/service_spec.cr
```diff
@@ -1045,6 +1045,30 @@ describe Autobot::Cron::Service do
...
+        schedule: Autobot::Cron::CronSchedule.new(kind: Autobot::Cron::ScheduleKind::Every, every_ms: 60000_i64),
+        kind: Autobot::Cron::PayloadKind::Exec,
+        command: "sleep 10"
+      )
+
+      service.run_job(job.id, force: true)
+
+      jobs = service.list_jobs
+      jobs.first.state.last_status.should eq(Autobot::Cron::JobStatus::Error)
+      jobs.first.state.last_error.to_s.should contain("124")
```
veelenga: This pins the made-up exit code. Once the error says what happened, assert on that instead, for example `should contain("timed out")`.

docs/cron.md
```diff
@@ -137,7 +137,7 @@ graph LR
     style SKIP fill:#78909c,stroke:#546e7a,color:#fff
 ```
 
-Exec commands run inside the sandbox (bubblewrap or Docker) by default. Set `sandbox: none` to run directly on the host.
+Exec commands run inside the sandbox (bubblewrap or Docker) by default. Set `sandbox: none` to run directly on the host. Execution is bounded by a 30-second timeout and output limits (up to 32 KB stored output) in both modes.
```
veelenga: Good addition, the 30s limit was not documented at all. Worth adding what the user sees when it hits: the job is marked as failed with a timed out error, and it runs again on its next schedule.

veelenga (commented):
Thanks, the problem in #157 is real and worth fixing. I checked it on `main` with `sandbox: none` and two jobs due every second:

| Setup | `echo hi` job after 6s |
|---|---|
| alone | ran, status `Ok` |
| next to a job running `sleep 600` | never ran |

`on_timer` runs due jobs one after another and re-arms the timer only when all of them are done, so one hung unsandboxed command stops every cron job until autobot restarts.

A few things I would change before this goes in (details inline):

1. **Report a timeout as a timeout.** `command exited with 124` makes up an exit code the command never returned. `main` now has `Tools::CommandRunner.run`, whose `Result` has `timed_out?`, and the sandboxed path already raises `command timed out after 30 seconds` and names the signal that killed a command. The direct path should give the same messages.
2. **One failure handler for both paths.** The same three lines are added twice.
3. **No new capture limit.** Stored output is cut to 32 KB right after the command returns.

One more note: the 30s limit is new for `sandbox: none`. Unsandboxed cron commands that run longer than that today will start to fail, so it is a behaviour change worth stating clearly in the PR description and the docs.

renich:
Thanks for the thorough review and empirical reproduction!

I have updated the branch with the suggested direction:
- Rebased cleanly onto latest `main`.
- Extended `Tools::CommandRunner.run` with `env: env` and `max_output_size : Int32? = DEFAULT_MAX_OUTPUT`, running direct exec via `CommandRunner.run` with `max_output_size: nil` so stored output is naturally bounded by the existing 32 KB `truncate_output` logic without an extra capture limit.
- Extracted `handle_exec_result` to unify outcome handling across both sandboxed and direct execution paths, reporting timeouts natively as explicit timeout events (`command timed out after X seconds`) instead of exit code 124.
- Added `cron.exec_timeout` (defaulting to 30s) to `CronConfig` and gateway setup so users can configure longer timeouts if needed.
- Added regression and schema specs in `spec/autobot/cron/service_spec.cr`, `spec/autobot/tools/command_runner_spec.cr`, and `spec/autobot/config/schema_spec.cr`.
- Updated documentation in `docs/cron.md` and `docs/configuration.md` clarifying the timeout behavior and configuration.

veelenga (commented):
Thanks for the update. Timeouts now read as timeouts, signal kills show the signal, both paths share `handle_exec_result`, and the config knob plus docs are in. I ran the touched specs, `crystal tool format --check` and ameba on the branch: all clean.

One real thing left, two small ones (details inline):

1. **Direct exec output is still unbounded**, while the title says it is bounded. `max_output_size: nil` turns the `CommandRunner` cap off, so everything the command writes is buffered and only then cut to 32 KB. Passing `MAX_STORED_OUTPUT_BYTES` instead fixes it, and the `Int32?` change in `CommandRunner` with its spec can then go.
2. The default `30` now lives in three places.
3. The new doc line gives the wrong reason for the 32 KB cap.

Also worth one explicit line in the PR description: unsandboxed commands that run longer than 30 s used to work and will now fail unless `cron.exec_timeout` is raised.

Aside, not for this PR: `autobot cron run <id>` builds `Cron::Service.new(store_path)` with defaults, so it ignores `cron.exec_timeout`, the same way it already ignores `tools.sandbox`. Follow-up material.

src/autobot/cron/service.cr
```diff
@@ -475,20 +472,21 @@ module Autobot
...
-        error = IO::Memory.new
-        status = Process.run(
-          "sh", {"-c", command},
-          output: output,
-          error: error,
+        result = Tools::CommandRunner.run(
+          "sh",
+          ["-c", command],
+          timeout: @exec_timeout.total_seconds.to_i,
+          max_output_size: nil,
```
veelenga: `nil` disables the cap, so the direct path buffers the whole output and `truncate_output` only cuts it afterwards. I checked with `sh -c "head -c 300000000 /dev/zero"`: 300 MB captured into a string, then thrown away. The 30 s timeout does not help here, a chatty command fills memory long before that.

```suggestion
          max_output_size: MAX_STORED_OUTPUT_BYTES,
```

`CommandRunner` then stops reading at 32 KB and `truncate_output` trims its notice and adds the cron one. The cron service spec passes with this change (57 examples, 0 failures). With it, the `Int32?` parameter in `CommandRunner` and the "does not truncate output when max_output_size is nil" spec are no longer needed, no caller wants unlimited output.

src/autobot/cli/gateway.cr
```diff
@@ -94,12 +94,14 @@ module Autobot
         end
 
         sandbox_config = config.tools.try(&.sandbox) || "auto"
+        exec_timeout = (config.cron.try(&.exec_timeout) || 30).seconds
```
veelenga: The default `30` is now in `CronConfig`, here and in `Service::EXEC_TIMEOUT`. The existing constant can serve as the fallback:

```suggestion
        exec_timeout = config.cron.try(&.exec_timeout.seconds) || Cron::Service::EXEC_TIMEOUT
```

docs/cron.md
```diff
@@ -139,6 +139,19 @@ graph LR
...
+Command execution is bounded by a 30-second timeout by default in both sandboxed and direct (`sandbox: none`) modes. When a command exceeds the timeout, it is terminated, the job is marked as failed with a `command timed out` error, and it runs again on its next scheduled interval.
+
+The timeout can be customized in `config.yml`:
+
+```yaml
+cron:
+  exec_timeout: 60  # seconds (default: 30)
+```
+
+Stored output is automatically capped at 32 KB per job (`PREV_OUTPUT`) to prevent runaway memory usage.
```
veelenga: The 32 KB cap exists so `PREV_OUTPUT` fits in the environment (E2BIG), not for memory, and sandboxed jobs are already cut at 10 KB by `Sandbox.exec` before that.

```suggestion
Output kept for `PREV_OUTPUT` is cut to 32 KB so it fits in the environment.
```

PR #160, opened 2026-09-21
fix(mcp): use single reader fiber and bound request timeouts

### LINKED ISSUE #159: MCP::Client leaks reader fibers and desynchronizes stdio stream on request timeout
### Problem Description
In `src/autobot/mcp/client.cr`, `read_response` spawns a background fiber to read lines from `@stdout` in a loop and send the parsed response through a channel:

```crystal
private def read_response(expected_id : Int64, timeout : Time::Span) : JSON::Any
  stdout = @stdout
  raise "MCP server not running" unless stdout

  channel = Channel(JSON::Any | Exception).new(1)

  spawn do
    begin
      loop do
        line = stdout.read_line
        parsed = JSON.parse(line)
        ...
        if response_id.as_i64? == expected_id
          channel.send(parsed)
          break
        end
      end
    rescue ex
      channel.send(ex)
    end
  end

  select
  when result = channel.receive
    ...
  when timeout(timeout)
    raise "MCP request timed out after #{timeout.total_seconds.to_i}s"
  end
end
```

### Root Cause
1. **Reader Fiber Leak**: When `timeout(timeout)` triggers, `read_response` raises immediately, but the spawned background fiber is neither cancelled nor aborted. It remains blocked on `stdout.read_line`.
2. **Stream Desynchronization**: The timed-out server process continues running. When subsequent requests are sent (e.g. request `id=2`), a new reader fiber is spawned. Both fibers now read from the same `stdout` stream concurrently. If the server eventually sends the delayed response for `id=1`, the old fiber or the new fiber may consume lines out of order, corrupting the JSON-RPC state machine.

### Expected Behavior
When an MCP request times out:
1. The client should stop the unresponsive server process and close its stdio pipes via `stop`.
2. Closing the pipes unblocks and terminates the orphaned reader fiber.
3. The server state is marked inactive (`alive? == false`), preventing stream desynchronization on subsequent calls.

renich:
When an MCP tool call timed out, the caller fiber stopped waiting while the reader fiber remained blocked waiting on `stdout` for that response ID. Subsequent calls spawned new fibers reading from the same stream, leading to stream desynchronization, response interleaving, and leaked fibers.

### Changes

- In `Mcp::Client#start`, spawn a single persistent reader fiber that parses lines and pushes valid response messages (`id` present, no `method`) to an unbuffered channel (`@responses`).
- In `read_response`, read from `@responses` until the monotonic deadline, discarding responses with stale IDs without terminating the server.
- Raise an explicit error (`MCP server closed the connection`) when the server closes the connection.
- Terminate the server process immediately if the initialization handshake fails or times out (`begin ... rescue ex; stop; raise ex; end`) to avoid leaked background processes.
- Add regression unit specs in `spec/autobot/mcp/client_spec.cr` covering slow tool recovery, handshake failure cleanup, and connection closure.
- Update `docs/mcp.md` with timeout and stream synchronization behavior.

Closes #159


src/autobot/mcp/client.cr
```diff
@@ -230,6 +234,7 @@ module Autobot
           end
           result
         when timeout(timeout)
+          stop
```
veelenga: This is the part I would not merge as is. There is no restart path, so one slow tool call turns into a permanent `MCP server '...' is not running` for every tool of this server (see the table in the review summary). On `main` the same server recovers on the next call.

It also runs inside `@mutex.synchronize`, and `stop` can wait up to `SHUTDOWN_GRACE` (2s), so the timeout error reaches the caller late and other callers stay blocked meanwhile.

With a single reader fiber a timeout needs no cleanup here at all.

src/autobot/mcp/client.cr
```diff
@@ -83,6 +84,11 @@ module Autobot
         process = @process
         return unless process
 
+        @stdin.try(&.close) rescue nil
+        @stdout.try(&.close) rescue nil
+        @stdin = nil
+        @stdout = nil
```
veelenga: Closing stdin before `SIGTERM` is a good change on its own and worth keeping: it matches the shutdown order in the MCP spec for stdio (close the input stream, wait, then `SIGTERM`, then `SIGKILL`), and many servers exit cleanly on EOF.

src/autobot/mcp/client.cr
```diff
@@ -42,6 +42,7 @@ module Autobot
         @command : String,
         @args : Array(String) = [] of String,
         @env : Hash(String, String) = {} of String => String,
+        @call_timeout : Time::Span = CALL_TIMEOUT,
```
veelenga, line 46: This option exists only so the spec can use a short timeout: config cannot set it, and `list_tools` still goes through the `CALL_TIMEOUT` default of `send_request`, so the client now has two sources for the same value. If the timeout should be injectable, it is cleaner to make it the single value used by every request.

spec/autobot/mcp/client_spec.cr
```diff
@@ -0,0 +1,31 @@
...
+    )
+
+    client.start
+    client.alive?.should be_true
+
+    expect_raises(Exception, /timed out/) do
+      client.call_tool("slow_tool", {} of String => JSON::Any)
+    end
+
+    client.alive?.should be_false
```
veelenga, line 57: This asserts the new behaviour (`alive?` is false) but not the bug from #159. The spec never sends a second request, so it would also pass if responses were still being taken by a stale reader. A regression spec for the issue would time out one call and then check that the next call gets its own answer, for both a late answer and no answer at all.

veelenga (commented):
Thanks for this, the bug in #159 is real. I reproduced it with a fake MCP server and ran the same probe on `main` and on this branch (client timeout 0.5s):

| Case | `main` | this PR |
|---|---|---|
| Slow call (answers after 1.5s), then 3 fast calls | times out, then all 3 work | times out, then all 3 fail with `MCP server not running` |
| Hung call (never answered), then 3 fast calls | all 3 time out, although the server answers them | all 3 fail at once with `MCP server not running` |
| Handshake answered with non-JSON | server process leaks | server process still leaks |

So the hung case gets better (fast failure instead of one full timeout per call), but the slow case gets worse. Nothing restarts a stopped client, so after one tool call over 60s `ProxyTool` answers `MCP server '...' is not running` until autobot is restarted, and every tool of that server is gone. On `main` that same server recovers on the next call.

### Suggested direction

The root cause is that `read_response` starts a new reader fiber for every request. After a timeout the old fiber stays blocked on `stdout` and takes the responses of later requests. Stopping the server removes that fiber, but the fiber is the thing to fix:

- Start **one** reader fiber per client in `start`. It parses each line and pushes responses (messages with an `id` and no `method`) into a channel.
- `read_response` receives from that channel until the deadline and skips ids it is not waiting for, so a late answer to a timed-out request is simply dropped.
- A closed channel means the server exited: raise `MCP server closed the connection`.

```crystal
private def read_response(expected_id : Int64, timeout : Time::Span) : JSON::Any
  responses = @responses
  raise "MCP server not running" unless responses

  deadline = Time.instant + timeout
  loop do
    select
    when response = responses.receive?
      raise "MCP server closed the connection" unless response
      return response if response["id"].as_i64? == expected_id
    when timeout(deadline - Time.instant)
      raise "MCP request timed out after #{timeout.total_seconds.to_i}s"
    end
  end
end
```

I prototyped this on `main`: with the same probe the slow case and the hung case both keep working after the timeout, no fiber is left behind, and the 36 existing MCP specs pass.

Separate but related: when the handshake fails for any reason, `start` raises and the caller drops the client, so the server process is never stopped. A `rescue ex; stop; raise ex` in `start` covers the timeout and every other handshake failure.

docs/mcp.md
```diff
@@ -107,6 +107,8 @@ echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":
 
 ### Timeouts
 
+If a request times out (30s for initialization handshake, 60s for tool calls), the client stops the server process and closes its stdio pipes to prevent stream desynchronization and orphaned background processes.
```
veelenga: If stop-on-timeout stays, the important part for users is missing here: the server is not restarted, so its tools stay unavailable until autobot restarts. With the single reader approach this paragraph would instead say that a timed-out request is dropped and the server keeps running.

renich:
Thanks for the review and prototype! The single persistent reader fiber with ID-based dispatch is a much better design than stopping the server on call timeouts.

I have updated the PR accordingly:
- Replaced per-request reader fibers with a single reader fiber spawned in `start` that parses incoming lines and pushes response messages (`id` present, no `method`) into `@responses`.
- `read_response` now loops on `@responses` until the monotonic deadline, returning matching response IDs and safely discarding late responses from timed-out calls so the server remains alive and usable.
- A closed channel raises `MCP server closed the connection`.
- Handshake is wrapped in `begin ... rescue ex; stop; raise ex; end` in `start`, preventing orphaned child processes if initialization times out or fails.
- Added regression unit tests in `spec/autobot/mcp/client_spec.cr` verifying slow call recovery with late response skipping, handshake failure process reaping, and remote connection closure.
- Updated `docs/mcp.md` to reflect the timeout and recovery semantics.

